PR #4, opened 2017-07-05
Make <snapd-glib/snapd-glib.h> include snapd-enum-types.h

jhenstridge:
The <snapd-glib/snapd-enum-types.h> header has guards to prevent it being included outside of <snapd-glib/snapd-glib.h>.  However, that header doesn't actually include it for you.


robert-ancell:
Thanks James!

PR #5, opened 2017-07-11
Fix snapd-qt build

aleixpol:
Don't define "signals", otherwise we had the following error:
/usr/include/glib-2.0/gio/gdbusintrospection.h:155:25: error: expected member name or ';' after declaration specifiers
  GDBusSignalInfo     **signals;
  ~~~~~~~~~~~~~~~       ^
/usr/include/QtCore/qobjectdefs.h:94:22: note: expanded from macro 'signals'
#     define signals Q_SIGNALS


robert-ancell:
Thanks Aleix!

PR #11, opened 2017-07-26
Only iterate our own GMainContext

iainlane:
See the commit message for my understanding of the issue. I admit to not being completely sure of the problem here, so it definitely needs review.

The user visible bug this is trying to fix is https://github.com/snapcore/snapd-glib/issues/10

### DIFF
--- a/snapd-glib/snapd-client.c
+++ b/snapd-glib/snapd-client.c
@@ -2537,14 +2537,14 @@ make_request (SnapdClient *client, RequestType request_type,
         request->auth_data = g_object_ref (priv->auth_data);
     request->client = client;
     request->request_type = request_type;
-    if (cancellable != NULL) {
-        request->cancellable = g_object_ref (cancellable);
-        request->cancelled_id = g_cancellable_connect (cancellable, G_CALLBACK (request_cancelled_cb), request, NULL);
-    }
     request->ready_callback = callback;
     request->ready_callback_data = user_data;
     request->progress_callback = progress_callback;
     request->progress_callback_data = progress_callback_data;
+    if (cancellable != NULL) {
+        request->cancellable = g_object_ref (cancellable);
+        request->cancelled_id = g_cancellable_connect (cancellable, G_CALLBACK (request_cancelled_cb), request, NULL);
+    }
     priv->requests = g_list_append (priv->requests, request);
 
     return request;
@@ -2572,6 +2572,7 @@ snapd_client_login_sync (SnapdClient *client,
                          GCancellable *cancellable, GError **error)
 {
     g_auto(SyncData) data = { 0 };
+    SnapdClientPrivate *priv;
 
     g_return_val_if_fail (SNAPD_IS_CLIENT (client), NULL);
     g_return_val_if_fail (username != NULL, NULL);
--- a/snapd-glib/snapd-client.c
+++ b/snapd-glib/snapd-client.c
@@ -2361,6 +2361,10 @@ sync_cb (GObject *object, GAsyncResult *result, gpointer user_data)
     SyncData *data = user_data;
     data->result = g_object_ref (result);
     g_main_loop_quit (data->loop);
+    /* If we run the loop now, we'll block forever, so get rid of it.
+     * For example, if we're passed an already cancelled GCancellable.
+     */
+    g_clear_pointer (&data->loop, g_main_loop_unref);
 }
 
 /**
@@ -2578,9 +2582,14 @@ snapd_client_login_sync (SnapdClient *client,
     g_return_val_if_fail (username != NULL, NULL);
     g_return_val_if_fail (password != NULL, NULL);
 
+    priv = snapd_client_get_instance_private (client);
+
     start_sync (client, &data);
+    g_main_context_push_thread_default (priv->context);
     snapd_client_login_async (client, username, password, otp, cancellable, sync_cb, &data);
-    g_main_loop_run (data.loop);
+    g_main_context_pop_thread_default (priv->context);
+    if (data.loop)
+        g_main_loop_run (data.loop);
     return snapd_client_login_finish (client, data.result, error);
 }
 
@@ -2724,12 +2733,18 @@ snapd_client_get_system_information_sync (SnapdClient *client,
                                           GCancellable *cancellable, GError **error)
 {
     g_auto(SyncData) data = { 0 };
+    SnapdClientPrivate *priv;
 
     g_return_val_if_fail (SNAPD_IS_CLIENT (client), NULL);
 
+    priv = snapd_client_get_instance_private (client);
+
     start_sync (client, &data);
+    g_main_context_push_thread_default (priv->context);
     snapd_client_get_system_information_async (client, cancellable, sync_cb, &data);
-    g_main_loop_run (data.loop);
+    g_main_context_pop_thread_default (priv->context);
+    if (data.loop)
+        g_main_loop_run (data.loop);
 
     return snapd_client_get_system_information_finish (client, data.result, error);
 }
@@ -2806,12 +2821,18 @@ snapd_client_list_one_sync (SnapdClient *client,
                             GCancellable *cancellable, GError **error)
 {
     g_auto(SyncData) data = { 0 };
+    SnapdClientPrivate *priv;
 
     g_return_val_if_fail (SNAPD_IS_CLIENT (client), NULL);
 
+    priv = snapd_client_get_instance_private (client);
+
     start_sync (client, &data);
+    g_main_context_push_thread_default (priv->context);
     snapd_client_list_one_async (client, name, cancellable, sync_cb, &data);
-    g_main_loop_run (data.loop);
+    g_main_context_pop_thread_default (priv->context);
+    if (data.loop)
+        g_main_loop_run (data.loop);
     return snapd_client_list_one_finish (client, data.result, error);
 }
 
@@ -2892,12 +2913,18 @@ snapd_client_get_icon_sync (SnapdClient *client,
                             GCancellable *cancellable, GError **error)
 {
     g_auto(SyncData) data = { 0 };
+    SnapdClientPrivate *priv;
 
     g_return_val_if_fail (SNAPD_IS_CLIENT (client), NULL);
 
+    priv = snapd_client_get_instance_private (client);
+
     start_sync (client, &data);
+    g_main_context_push_thread_default (priv->context);
     snapd_client_get_icon_async (client, name, cancellable, sync_cb, &data);
-    g_main_loop_run (data.loop);
+    g_main_context_pop_thread_default (priv->context);
+    if (data.loop)
+        g_main_loop_run (data.loop);
     return snapd_client_get_icon_finish (client, data.result, error);
 }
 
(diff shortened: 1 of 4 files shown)


robert-ancell:
You should definitely merge in 4bdc4a41e3a1a34a9ce842d67b36bd246d9c07e3 as that's a clear stuff-up on my part.

robert-ancell:
As far as I can tell we're not iterating the main context (since 4bdc4a41e3a1a34a9ce842d67b36bd246d9c07e3) so I'm not sure what part of the docs we're not doing correctly.

I really, really wanted to avoid using threads because it seems unnecessary and a likely cause of similar obscure issues. But glib seems to want to push you that way :(

I hadn't ready about using g_main_context_invoke. That seems worthwhile to do as a separate commit. Hopefully then the remaining changes will be clearer to understand.

iainlane:
On Thu, Jul 27, 2017 at 06:54:51AM -0700, Robert Ancell wrote:
> You should definitely merge in 4bdc4a41e3a1a34a9ce842d67b36bd246d9c07e3 as that's a clear stuff-up on my part.

I don't have commit access - if you like, you could do that for me.

(I just noticed that I should have used g_source_destroy too, since
g_source_remove only works for the default main context ...

switched to that - the commit is now
558d38d9d2ed850fc86da857cfacf94971227f07)

Cheers,

-- 
Iain Lane                                  [ iain@orangesquash.org.uk ]
Debian Developer                                   [ laney@debian.org ]
Ubuntu Developer                                   [ laney@ubuntu.com ]

snapd-glib/snapd-client.c
```diff
@@ -2443,16 +2462,22 @@ snapd_client_connect_sync (SnapdClient *client,
...
-    GTask *task;
-    g_autoptr(GError) error = NULL;
+    SnapdClientPrivate *priv;
+    g_autoptr(GTask) task = NULL;
+
+    g_return_if_fail (SNAPD_IS_CLIENT (client));
+
+    priv = snapd_client_get_instance_private (client);
 
+    g_main_context_push_thread_default (priv->context);
```
iainlane: Here, the task's callback will be called in the thread-default main context, so we need to change it.

robert-ancell:
I think you have to be in the snapcore team and @niemeyer is the only one who can add you?

snapd-glib/snapd-client.c
```diff
@@ -2578,7 +2605,7 @@ snapd_client_login_sync (SnapdClient *client,
     g_return_val_if_fail (username != NULL, NULL);
     g_return_val_if_fail (password != NULL, NULL);
 
-    data.loop = g_main_loop_new (NULL, FALSE);
```
iainlane: This one is necessary too I think it was just an omission.

snapd-glib/snapd-client.c
```diff
@@ -4168,15 +4198,20 @@ snapd_client_install_stream_async (SnapdClient *client,
...
     g_return_if_fail (SNAPD_IS_CLIENT (client));
     g_return_if_fail (G_IS_INPUT_STREAM (stream));
 
+    priv = snapd_client_get_instance_private (client);
+
     request = make_request (client, SNAPD_REQUEST_INSTALL_STREAM, progress_callback, progress_callback_data, cancellable, callback, user_data);
     request->install_flags = flags;
     request->snap_stream = g_object_ref (stream);
     request->snap_contents = g_byte_array_new ();
+    g_main_context_push_thread_default (priv->context);
```
iainlane: Same as for GTask above.

snapd-glib/snapd-client.c
```diff
@@ -2443,16 +2462,22 @@ snapd_client_connect_sync (SnapdClient *client,
...
 
+    g_main_context_push_thread_default (priv->context);
     task = g_task_new (client, cancellable, callback, user_data);
-    if (snapd_client_connect_sync (client, cancellable, &error))
-        g_task_return_boolean (task, TRUE);
-    else
-        g_task_return_error (task, error);
+    g_main_context_pop_thread_default (priv->context);
+
+    g_task_run_in_thread (task, do_connect_thread);
```
iainlane: This bit is because the operation wasn't really async before - I guess it could be a separate commit.

iainlane:
I just made a review to be able to highlight some places where changes are still necessary even after that commit. Some operations happen in the thread-default main context.

It's going to have to be something to continue to care about on an ongoing basis, but the good thing is - since we no longer ever iterate the thread-default main context, the testsuite now hangs if you get this wrong, since operations scheduled on it will never be processed (unless something else runs it).

snapd-glib/snapd-client.c
robert-ancell: Yep, that one must have got missed when I modified the other ones.

snapd-glib/snapd-client.c
robert-ancell: snapd_client_connect_async is a lie - I added it there for completeness. But I think it's still safe the way it was, no?

iainlane: I don't *think* there's a problem in the old way. If you want, I'll revert this bit and we can maybe look at it later. There's a bit of an overhead in doing it all sync, so it might be worth thinking about, but I can see that it's probably not necessary for this PR.

snapd-glib/snapd-client.c
```diff
@@ -2714,6 +2722,7 @@ snapd_client_get_auth_data (SnapdClient *client)
  *
  * Since: 1.0
  */
+
```
robert-ancell: Whitespace added, please remove..

snapd-glib/snapd-client.c
```diff
@@ -2466,6 +2472,8 @@ snapd_client_connect_async (SnapdClient *client,
 gboolean
 snapd_client_connect_finish (SnapdClient *client, GAsyncResult *result, GError **error)
 {
+    g_return_val_if_fail (g_task_is_valid (result, client), FALSE);
```
robert-ancell: I fixed this on master in the same style of the other _finish methods.

snapd-glib/snapd-client.c
```diff
@@ -2417,6 +2415,7 @@ snapd_client_connect_sync (SnapdClient *client,
 
     if (priv->read_source == NULL) {
         priv->read_source = g_socket_create_source (priv->snapd_socket, G_IO_IN, NULL);
+        g_source_set_name (priv->read_source, "snapd-glib-read-source");
```
robert-ancell: Applied upstream as a single commit.

snapd-glib/snapd-client.c
```diff
@@ -5843,17 +5857,19 @@ snapd_client_finalize (GObject *object)
...
-    if (priv->read_source != NULL)
-        g_source_destroy (priv->read_source);
-    g_clear_pointer (&priv->read_source, g_source_unref);
     g_socket_close (priv->snapd_socket, NULL);
     g_clear_object (&priv->snapd_socket);
     g_clear_pointer (&priv->buffer, g_byte_array_unref);
+
+    if (priv->read_source != NULL)
+        g_source_destroy (priv->read_source);
+    g_clear_pointer (&priv->read_source, g_source_unref);
```
robert-ancell: Why did the read source destruction get moved to last? Is this an existing bug?

iainlane:
Sorry, I bloated the diff again. I think it's in more or less logical commits, but maybe there's more to do still.

The problem was that I was pushing the default main context when processing an async operation. This would have meant that its callback would be called in our private main context, which is not where it should be called. I think instead that we should push the default main context to set our private one whenever we perform a sync operation. For most of them this probably doesn't actually matter - but if you do it systematically then it reduces the chance of mistakes.

What do you think? Maybe we can abstract this boilerplate somehow.

robert-ancell:
I merged into master all the small fixes in this branch, so once rebased it's much easier to see what the proposed changes are.

This is effectively the change you are proposing:

```
diff --git a/snapd-glib/snapd-client.c b/snapd-glib/snapd-client.c
index ef08a91..5327f2d 100644
--- a/snapd-glib/snapd-client.c
+++ b/snapd-glib/snapd-client.c
@@ -5873,7 +5873,7 @@ snapd_client_init (SnapdClient *client)
 {
     SnapdClientPrivate *priv = snapd_client_get_instance_private (client);
 
-    priv->context = g_main_context_ref_thread_default ();
+    priv->context = g_main_context_new ();
     priv->user_agent = g_strdup ("snapd-glib/" VERSION);
     priv->buffer = g_byte_array_new ();
 }
```

This would have the effect of making the socket read event happen on a context that was only used within snapd-glib. Any async call into the library would then fail because no main loop would ever poll for the read event.

The programming guidelines you reference state:

>   always document which context each callback will be dispatched in. For example, “callbacks will always be dispatched in the context which is the thread-default at the time of the object’s construction”.

Which is the current behaviour AFAICT.

robert-ancell:
I've added some async test cases to master which might help with this branch.

iainlane:
On Mon, Aug 14, 2017 at 02:09:54AM +0000, Robert Ancell wrote:
> This would have the effect of making the socket read event happen on a context that was only used within snapd-glib. Any async call into the library would then fail because no main loop would ever poll for the read event.
> 
> The programming guidelines you reference state:
> 
> >   always document which context each callback will be dispatched in. For example, “callbacks will always be dispatched in the context which is the thread-default at the time of the object’s construction”.
> 
> Which is the current behaviour AFAICT.

This is the thing you violate currently

> Never iterate a context created outside the library, including the
> global-default or thread-default contexts. Otherwise, GSources created
> in the application may be dispatched when the application is not
> expecting it, causing re-entrancy problems for the application code. 

if you just ref the existing thread default main context and iterate it.

asynchronous operations should have their callbacks called in *their*
thread-default main context, not the private one we're creating here.
snapd-glib's own internal operations - including sync calls - should use
the private one.

That is to say, the change you're suggesting is wrong isn't wrong. If
there's an error then it will be around deciding whether priv->context
or the thread default context is used at some point.

Do you want to close this PR and take care of it yourself?

Cheers,

-- 
Iain Lane                                  [ iain@orangesquash.org.uk ]
Debian Developer                                   [ laney@debian.org ]
Ubuntu Developer                                   [ laney@ubuntu.com ]

robert-ancell:
I thought a lot about this more today and I think the proposed solution will solve the sync calls but break the async ones. I get your point about iterating the external context (i.e. in the sync calls).

I'll update this PR / try another method and see if that works in all cases.

Thanks!

iainlane:
On Tue, Aug 15, 2017 at 04:59:34AM +0000, Robert Ancell wrote:
> I thought a lot about this more today and I think the proposed solution will solve the sync calls but break the async ones. I get your point about iterating the external context (i.e. in the sync calls).

Iterating the context happens in the async calls too, for the read_cb.
Off the top of my head I guess the async problem is that read_cb is
attached to priv->context and nothing is running that for asynchronous
calls. Also we invoke respond_cb in priv->context too; that might be an
issue - IIRC that is what invokes the user's callback.

It might make life easier to port to GTask.

-- 
Iain Lane                                  [ iain@orangesquash.org.uk ]
Debian Developer                                   [ laney@debian.org ]
Ubuntu Developer                                   [ laney@ubuntu.com ]

robert-ancell:
Ugh, every time I try to get more of this fix in something else breaks...

What I'm working on:

1. Ensuring that the callbacks are sent back in the context of the code that called the method (landed).

2. Getting your threaded mock snapd change landed, because the context inside that is never going to be used when we use sync calls that create their own context. This is a little tricky because the test code access a few variables set inside the mock code and will have to have a mutex around that. This is currently breaking in the GCancellable tests for some reason.

3. Replacing the read_cb source with one that is created for every request. I *think* this is the right way to do this because then it's always attached to the right context. It should be OK to have multiple sources on the same socket as each response will be handled in the right context after change 1. I wanted to handle responses completely asynchronously, but I don't think there's a requirement from snapd to do this at the moment.

robert-ancell:
I've got those fixes all in now, but I'm now getting some occasional crashes on startup when running with gnome-software...

iainlane:
Should we close this PR? If you get a trace you can open a bug and I'll help look if you want.

iainlane:
Oh, well, I said "should" and then pressed the close button. :)

PR #14, opened 2017-08-24
Add API to control whether polkit authorisation checks allow interaction

jhenstridge:
This branch is intended to complement https://github.com/snapcore/snapd/pull/3795 by adding `set_allow_interaction` and `get_allow_interaction` APIs to `SnapdClient`.  The idea is to let the client decide whether snapd's polkit authorisation checks should allow interaction or not.

One question is what the default should be for snapd-glib.  At present I've set it to default to allow interaction on the assumption that most users of the library will be graphical.  But it would be pretty easy to disallow interaction by default.

The change should be backward compatible with old versions of snapd: they will simply ignore the extra request header.

### DIFF
--- a/snapd-glib/snapd-client.c
+++ b/snapd-glib/snapd-client.c
@@ -85,6 +85,9 @@ typedef struct
     GMutex requests_mutex;
     GList *requests;
 
+    /* Whether to send the X-Allow-Interaction request header */
+    gboolean allow_interaction;
+
     /* Data received from snapd */
     GMutex buffer_mutex;
     GByteArray *buffer;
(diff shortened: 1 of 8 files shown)


robert-ancell:
Looks good to me, happy to merge once snapd approves the use of this field.

robert-ancell:
I brought over some of your test case changes into master, so this branch will have to be rebased again.

jhenstridge:
I've updated the branch, also changing the value we set the request header to, as requested in the snapd PR.

robert-ancell:
I squashed the commits and fixed up some documentation issues in master for this, thanks!

PR #15, opened 2017-08-24
configure: json-glib 1.1 is enough to build

morphis:
json-glib 1.2 is not available on Ubuntu 16.04 (xenial) but 1.1.2 is. A build
with 1.1.2 goes fine so lets use that as new baseline.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -29,7 +29,7 @@ dnl === Dependencies ==========================================================
 
 GLIB_REQUIRED=2.46
 LIBSOUP_REQUIRED=2.32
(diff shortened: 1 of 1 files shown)


robert-ancell:
The reason I used 1.2 was:

/**
 * json_node_ref:
 * @node: a #JsonNode
 *
 * Increment the reference count of @node.
 *
 * Since: 1.2
 * Returns: (transfer full): a pointer to @node
 */

robert-ancell:
Ah, I see, the devel version had this. Thanks!

PR #16, opened 2017-09-07
Add support for channels

robert-ancell:


robert-ancell:
@aleixpol can you help me solve a Qt issue with this merge...

There's an enum (QSnapdConfinement) that was used in QSnapdSnap. However, now it needs to be used in both QSnapdSnap and QSnapdChannel. Is there a way I can make the headers work with this?

Perhaps I need to move the enums out of the QSnapdSnap class?

aleixpol:
It would be a possibility. There's something I don't understand though: A snap has a QSnapdConfinement and N channels, and each of these has a QSnapdConfinement?

Is `snap.confinement` a shortcut to `snap.currentChannel.confinement`?
If so maybe a possibility would be to just move it into QSnapdChannel (yes, API-incompatible change).

Otherwise, you can move it into the channel, in the end QSnapdSnap will need to know about QSnapdChannel but not the other way around.

robert-ancell:
I've been asking the architects here:
https://forum.snapcraft.io/t/snapd-channel-information/2004

If you do a list () then confinement would be the confinement that the snap is installed with, if you do a find () then the confinement would not be useful if the client understands channels.

None of the enums are specifically connected to classes (and I don't know how the snapd API will change in the future). Is there a better convention to move them elsewhere (e.g. a SnapdEnum class that contains all the enums?)

aleixpol:
The reason why I'm not suggesting to move it elsewhere is because I think you wanted to make this library usable on QML. For this case, exposing external enums to QML is a bit weird.

PR #22, opened 2017-09-26
Move Qt enums to an enum class and rename to avoid collisions

robert-ancell:


robert-ancell:
@aleixpol can you do a quick review to see if this makes sense?

This will be an API break, since the enums have changed names and values.
I don't think we can remove the enums, since they make up the Qt ABI.
I'd like to mark them as deprecated in some way, but I don't know how to do that in Qt.

aleixpol:
To mark something as deprecated you can use `QT_DEPRECATED`.

FWIW, it's not ABI incompatible, but ABI. So things should still work but it won't compile anymore.

We haven't released any Discover with snapd-glib dependency just yet, your call.

PR #23, opened 2017-09-26
Fix runtime assertion when snapd socket is NULL

ikeydoherty:
This was found whilst using the Python GIR bindings and deliberately
causing the snapd.socket service to fail (i.e. emulation of a missing
snapd service at runtime). During process exit and finalisation of the
SnapdClient object, the socket would be unconditionally closed regardless
of whether connection had succeeded or not.

This commit adds a very trivial guard around the cleanup.

Signed-off-by: Ikey Doherty <ikey@solus-project.com>


robert-ancell:
Thanks Ikey!

ikeydoherty:
np - cheers for merge

PR #26, opened 2017-10-25
tests: port MockSnapd to SoupServer

jhenstridge:
I was noticing some instability in the test suite when working on another change, and did a bit of refactoring of the MockSnapd code to see if it improved matters.  Unfortunately it seems I was in the middle of some work you were doing on master, so some of this might not be appropriate any more.  I thought I'd submit the PR all the same.

The main changes in this branch are:

1. directly use SoupServer rather than reimplementing an HTTP server
2. push some of the initialisation and clean up to the worker thread, so quitting the worker thread's main loop cleans things up.

### DIFF
--- a/tests/mock-snapd.c
+++ b/tests/mock-snapd.c
@@ -23,15 +23,6 @@
 
 typedef struct
 {
-    MockSnapd *snapd;
-    GSocket *socket;
-    GSource *read_source;
-    GByteArray *buffer;
-    gsize n_read;
-} MockClient;
-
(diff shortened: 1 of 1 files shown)


robert-ancell:
If you can rebase this then I'd love to use it. I wasn't aware you could use SoupServer in this way.

jhenstridge:
I've ported the changes forward to take into account the `mock_snapd_start` changes (including forwarding GErrors from the worker thread back to the main thread.

PR #27, opened 2017-10-25
snapd-glib: Use response status code to set error codes as a backup.

jhenstridge:
If JSON error response doesn't include a "kind" field, fall back on the HTTP response status code.

### DIFF
--- a/snapd-glib/snapd-json.c
+++ b/snapd-glib/snapd-json.c
@@ -246,16 +246,39 @@ parse_error_response (JsonObject *root, GError **error)
                              SNAPD_ERROR,
                              SNAPD_ERROR_NEEDS_CLASSIC_SYSTEM,
                              message);
-    else if (status_code == SOUP_STATUS_BAD_REQUEST)
-        g_set_error_literal (error,
-                             SNAPD_ERROR,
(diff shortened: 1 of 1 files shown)


snapd-glib/snapd-json.c
```diff
@@ -165,14 +165,12 @@ snapd_json_get_date_time (JsonObject *object, const gchar *name)
...
 static void
-parse_error_response (JsonObject *root, GError **error)
+parse_error_response (int status_code, JsonObject *root, GError **error)
 {
     const gchar *kind = NULL, *message = NULL;
-    gint64 status_code;
     JsonObject *result;
 
     result = snapd_json_get_object (root, "result");
-    status_code = snapd_json_get_int (root, "status-code", 0);
```
robert-ancell: Did you stop using the status code as provided in the JSON for a reason? I used this one over the HTTP one because I think the intention upstream is for the JSON responses to be potentially sent using other methods (e.g. websockets).

jhenstridge:
I've switched back to using the status-code from the JSON message.  In general, the two should be the same, but the HTTP status code is authoritative.  It isn't really important to this PR though, so I've backed that change out.

PR #31, opened 2018-01-25
Fix crash when calling snapd_login_async

azzar1:
snapd_login_async right now is broken in several ways:
1) request is not passed around
2) the callback is not  called in case of errors
3) the error is not properly freed
4) request struct is not properly initialized


robert-ancell:
Jeesh, embarrassing... Commited with a small whitespace fix and added some tests. Thanks!

PR #34, opened 2018-04-03
Properly unref SnapdClient

azzar1:
Make sure SnapdClient reference is properly handled once a request has completed or cancelled. Also feeks few memory leaks in snapd-qt and in the tests.

https://bugs.launchpad.net/ubuntu/+source/snapd-glib/+bug/1760795


snapd-qt/assertion.cpp
```diff
@@ -19,26 +19,32 @@ QSnapdAssertion::QSnapdAssertion (const QString& contents, QObject *parent) :
...
 {
-    g_auto(GStrv) headers = NULL;
+    GStrv headers = NULL;
     QStringList result;
 
     headers = snapd_assertion_get_headers (SNAPD_ASSERTION (wrapped_object));
     for (int i = 0; headers[i] != NULL; i++)
         result.append (headers[i]);
+
+    g_strfreev (headers);
```
robert-ancell: This change is exactly what was done before with `g_auto(GStrv)`...

snapd-glib/requests/snapd-request.c
```diff
@@ -61,7 +61,7 @@ void
 _snapd_request_set_source_object (SnapdRequest *request, GObject *object)
 {
     SnapdRequestPrivate *priv = snapd_request_get_instance_private (SNAPD_REQUEST (request));  
-    priv->source_object = g_object_ref (object);
+    priv->source_object = object ? g_object_ref (object) :  NULL;
```
robert-ancell: This doesn't seem useful - object will always be non-null in this case.

robert-ancell:
I've committed some of the simple changes directly - please rebase.

Aside from that there seem to be a couple of issues raised:
1. The reference to SnapdClient in SnapdRequest. I'm not sure if this should be a strong or a weak reference. It's only used in the callback and it's been a real headache why it exists in the AsyncResult interface at all...
2. The cancel deadlock - can you explain what is happening here?

snapd-qt/assertion.cpp
azzar1: Ops. You're right. I guess it's a leftover. Removing!

snapd-glib/requests/snapd-request.c
azzar1: Just wanted to be avoid problems that could arise in the future.

azzar1:
Thanks for the changes!

1. I guess it depends on how we wants our API to work. If we switch to weak reference that a _SnapdRequest_ will not be completed if the _SnapdClient_ is unreferenced before. In contrast,  if we keep a strong reference to _SnapdClient_, the _SnapdRequest_ will be completed unless we cancel it.

2.  At the moment when a request is cancelled the following function will be called `snapd_request_return`. This will not unref the SnapRequest and free the resources allocated by `RequestData`. To do so we must call `snapd_request_complete`. To conclude we will end-up calling g_cancellable_disconnect from the "cancelled" signal handler and this causes a deadlock as documented [here](https://developer.gnome.org/gio/stable/GCancellable.html#g-cancellable-disconnect).

These fix have to go together otherwise tests will fail.

robert-ancell:
OK, I think I've landed all these changes into master. I had to fix up the idle callback - it was passing the data through but that could have been freed.

PR #36, opened 2018-06-08
Support sandbox-features

robert-ancell:


aleixpol:
Hey, just looked at it, I'm not super sure what you're aiming to achieve, can I have some context?
What will this QHash offer?

zyga:
This is nice. How do you plan on using this in the future?

robert-ancell:
@aleixpol sorry for the delay (was travelling). This is currently on the snap command line as this:

```
$ snap debug sandbox-features 
apparmor:             kernel:caps kernel:dbus kernel:domain kernel:file kernel:mount kernel:namespaces kernel:network kernel:policy kernel:ptrace kernel:query kernel:rlimit kernel:signal
confinement-options:  classic devmode strict
dbus:                 mediated-bus-access
kmod:                 mediated-modprobe
mount:                freezer-cgroup-v1 layouts-beta mount-namespace per-snap-persistency per-snap-profiles per-snap-updates per-snap-user-profiles stale-base-invalidation
seccomp:              bpf-argument-filtering kernel:allow kernel:errno kernel:kill_process kernel:kill_thread kernel:log kernel:trace kernel:trap
udev:                 device-cgroup-v1 tagging
```

And we are exposing it via snapd-glib. There's an arbitrary number of backends (so I stored them in a QHash) and each backend has n strings that describe features that are supported.

@zyga the particular feature we want to use this for is the `confinement-option` - that tells you what snap confinements the system supports. This was requested for Fedora where we wanted to be sure to only mark snaps as sandboxed if both the snap is in strict mode and the system supports it.

zyga:
@robert-ancell I see, nice, this makes sense.

snapd-glib/snapd-system-information.c
```diff
@@ -219,6 +221,23 @@ snapd_system_information_get_os_version (SnapdSystemInformation *system_informat
     return system_information->os_version;
 }
 
+/**
+ * snapd_system_information_get_sandbox_features:
+ * @system_information: a #SnapdSystemInformation.
+ *
+ * Gets the sandbox features that snapd has.
```
aleixpol: "that snapd has"? Maybe "that snapd requires"?

It should also document what the key's value is.

robert-ancell:
@aleixpol yes, it's just querying the capabilities. Understanding the values requires knowledge of how snapd operates. I've updated the docstring, thanks!

PR #41, opened 2018-08-21
Support new auth-cancelled error

robert-ancell:


robert-ancell:
Associated snap PR https://github.com/snapcore/snapd/pull/5690

PR #42, opened 2018-09-20
Reconnect to snapd if disconnected while trying to send the request

jhenstridge:
If the remote end has disconnected our socket connection, we will only see the error when we next try to write a request to snapd.  If we see this error, we should reconnect and try again.

I've also added a boolean flag to prevent us looping endlessly if snapd is misbehaving.

### DIFF
--- a/snapd-glib/snapd-client.c
+++ b/snapd-glib/snapd-client.c
@@ -793,6 +793,7 @@ send_request (SnapdClient *client, SnapdRequest *request)
     g_byte_array_append (request_data, (const guint8 *) buffer->data, buffer->length);
 
     for (retry = 0; retry < 2; retry++) {
+        g_clear_error (&local_error);
         if (data->read_source != NULL) {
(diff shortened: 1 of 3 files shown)


snapd-glib/snapd-client.c
```diff
@@ -791,44 +792,60 @@ send_request (SnapdClient *client, SnapdRequest *request)
     buffer = soup_message_body_flatten (message->request_body);
     g_byte_array_append (request_data, (const guint8 *) buffer->data, buffer->length);
 
-    if (priv->snapd_socket == NULL) {
-        g_autoptr(GSocketAddress) address = NULL;
-        g_autoptr(GError) error_local = NULL;
+    for (retry = 0; retry < 2; retry++) {
```
robert-ancell: Why is it hard-coded to retry only twice?

robert-ancell:
I reworked this patch and committed it to master, thanks!

PR #46, opened 2018-10-29
Qt: Mark constant properties as such

aleixpol:
Otherwise we get a warning from QML every time it's accessed that it may change.


robert-ancell:
Thanks @aleixpol !

PR #48, opened 2018-11-07
Ensure a request is completed only once.

azzar1:
The function snapd_request_complete is called twice for the same SnapdRequest in
case the request is cancelled after a response message has already been received
and parsed. In order to avoid double refs we need to enforce that a request
cannot be completed twice.

Fixes: https://bugs.launchpad.net/bugs/1798360

### DIFF
--- a/snapd-glib/requests/snapd-request.c
+++ b/snapd-glib/requests/snapd-request.c
@@ -87,14 +87,14 @@ respond_cb (gpointer user_data)
     return G_SOURCE_REMOVE;
 }
 
-void
+gboolean
 _snapd_request_return (SnapdRequest *request, GError *error)
 {
     g_autoptr(GSource) source = NULL;
     SnapdRequestPrivate *priv = snapd_request_get_instance_private (request);
 
     if (priv->responded)
-        return;
+        return FALSE;
     priv->responded = TRUE;
     if (error != NULL)
         priv->error = g_error_copy (error);
@@ -102,6 +102,8 @@ _snapd_request_return (SnapdRequest *request, GError *error)
     source = g_idle_source_new ();
     g_source_set_callback (source, respond_cb, g_object_ref (request), g_object_unref);
     g_source_attach (source, _snapd_request_get_context (request));
+
+    return TRUE;
 }
 
 gboolean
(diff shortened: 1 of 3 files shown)


robert-ancell:
Can you reproduce this with a test? I'd like to add a test with this change so it doesn't happen again. I'm also not sure if the fix is in the right place - it seems to be fixing the symptom, not the cause of the problem.

azzar1:
> Can you reproduce this with a test? I'd like to add a test with this change so it doesn't happen again.

```c
#include <glib.h>
#include <snapd-glib/snapd-glib.h>

SnapdClient *client;

static void sync_cb (GObject *object, GAsyncResult *result, gpointer user_data)
{}

static gboolean idle_cb (gpointer user_data) {
        g_autoptr(GCancellable) cancellable = g_cancellable_new ();

        snapd_client_find_refreshable_async (client, cancellable, sync_cb, NULL);
        g_cancellable_cancel (cancellable);

       return TRUE;
}

int main() {
        g_autoptr(GMainLoop) loop;

        client = snapd_client_new ();
        g_timeout_add (10, idle_cb, NULL);

        loop = g_main_loop_new (NULL, FALSE);
        g_main_loop_run (loop);
}
```

This should be enough to make it crash or to trigger the invalid read message.

> I'm also not sure if the fix is in the right place - it seems to be fixing the symptom, not the cause of the problem.

I do agree. Another solution I tested involved keeping tracking of the source idle used when a request is cancelled, making sure that if the idle source is not NULL, a call to snapd_complete_request would be ignored. But this solution too was not fixing the cause but the symptom.

The true issue here is the reference cycles that there are between RequestData and data. If we want to fix it properly we need to get rid of those.

robert-ancell:
> The true issue here is the reference cycles that there are between RequestData and data. If we want to fix it properly we need to get rid of those.

Patches welcome :)

robert-ancell:
I'm going to think this one over on the weekend and promise to land or modify by Monday :)

robert-ancell:
The root cause of this was the request being unreffed when removed from a GList. However, if the request wasn't in the GList then a double unref occurred. This is fixed in bd8b77d7ab20e5d49427f4d3d5ee34b4d49fa4a5. It now passes your test without error - please let me know if you think there doesn't fix the problem as your patch proposed.

Thanks for tracking this one down @azzar1!

azzar1:
> The root cause of this was the request being unreffed when removed from a GList. However, if the request wasn't in the GList then a double unref occurred. This is fixed in bd8b77d. It now passes your test without error - please let me know if you think there doesn't fix the problem as your patch proposed

This too seems to be fixing the symptom but it should work anyways. Thanks!

PR #58, opened 2019-03-24
Ensure that the C compiler operates in C99 mode

Conan-Kudo:
The system GCC version in Red Hat Enterprise Linux 7 is GCC 4.8, which predates the switch to C99 by default for compiling C code, which happened in GCC 5.

This ensures that we detect the C compiler correctly and set it to operate in C99 mode.

Signed-off-by: Neal Gompa <ngompa13@gmail.com>


zyga:
There are two errors:
```
ERROR: test-qt - missing test plan
ERROR: test-qt - exited with status 127 (command not found?)
```

Conan-Kudo:
@zyga I didn't do anything to influence those errors, though?

robert-ancell:
@Conan-Kudo right, that's an existing error.

PR #59, opened 2019-03-31
Allow searching via common-id

robert-ancell:


robert-ancell:
snapd change: https://github.com/snapcore/snapd/pull/6635

PR #60, opened 2019-04-14
client: Fix snapd_client_set_socket_path

azzar1:
Properly reset socket_path to the default if NULL is passed to
snapd_client_set_socket_path. Also add unit test.


robert-ancell:
Good catch, thanks @azzar1 !

PR #64, opened 2019-05-03
Support new /v2/interfaces&select= API, deprecate old API

robert-ancell:


jhenstridge:
Looks good!

jhenstridge:
I missed the deprecations part: I don't think that part is correct.  The new `/v2/interfaces?select=` API end point does not provide all the information that the old one does, so the deprecations are probably premature.

The newer API will tell you about interfaces supported by snapd along with which snaps provide plugs and slots using those interfaces, but it offers no information about connections between those plugs and slots.  So with this PR, you'd be forced to use deprecated API to check connection state.

The new API could probably be extended to cover connections, but it doesn't currently provide that.  It'd be really useful if it did, since the old API doesn't provide any server side filtering.

robert-ancell:
Another commit added /v2/connections which replaces the old interfaces API.

PR #80, opened 2019-12-12
Remove screenshot decoding code

robert-ancell:
snapd hasn't populated these since 2.40 and instead puts in a note that is
confusing snapd-glib into thinking there is one screenshot.

snapd-glib will now always return 0 screenshots.


robert-ancell:
See https://forum.snapcraft.io/t/screenshots-are-going-away/8086 for rationale in snapd.

PR #82, opened 2020-02-27
Make the build reproducible

lamby:
Whilst working on the [Reproducible Builds](https://reproducible-builds.org/) effort we noticed that snapd-glib could not be built reproducibly. This is because it embedded the full build path in comments.

This was originally filed in Debian as [#952694](https://bugs.debian.org/952694).


robert-ancell:
Thanks @lamby!

PR #84, opened 2020-04-02
Add snapd_interface_make_label

robert-ancell:
cc @kenvandine


robert-ancell:
The purpose of this is to remove this code from gnome-control-center and snap-store and store it in a central location. We may still consider snapd to be a better location for this longer term - new interfaces will need to be added to snapd-glib for this to keep up to date. This would probably be easier in snapd once translatable metadata is available.

PR #93, opened 2020-12-01
Do not leak the d-pointer of the class

aleixpol:


robert-ancell:
Thanks! I guess I should be using QScopedPointer in all the other cases too?

aleixpol:
> Thanks! I guess I should be using QScopedPointer in all the other cases too?

Yes, either that or `delete d_ptr` on the destructor. I didn't see other leaks though.

PR #94, opened 2020-12-01
Use QScopedPtr for d_ptr

robert-ancell:


robert-ancell:
@aleixpol is this OK. Will it have any ABI impacts?

aleixpol:
Looks good to me. ABI shouldn't be affected because sizeof(QScopedPointer) is the same as sizeof(void*).

PR #95, opened 2021-02-09
snapd-glib: add error_value out parameter to _snapd_json_parse_response

jhenstridge:
This PR modifies the _snapd_json_parse_response` utility function to add a `JsonNode **error_value` out parameter that will be set to the `value` property of error responses it decodes.

The eventual goal is to update the `snapd_client_run_snapctl*` functions to correctly support non-zero exit codes, which are returned by snapd as error responses with the exit code, and command output stored in the error value.  I've submitted this PR on its own since it touches every request handler class, and didn't want to mix in a largely mechanical changes with ones that deserve more scrutiny.

### DIFF
--- a/snapd-glib/requests/snapd-json.c
+++ b/snapd-glib/requests/snapd-json.c
@@ -176,12 +176,17 @@ _snapd_json_get_date_time (JsonObject *object, const gchar *name)
 }
 
 static void
-parse_error_response (JsonObject *root, GError **error)
+parse_error_response (JsonObject *root, JsonNode **error_value, GError **error)
 {
     JsonObject *result = _snapd_json_get_object (root, "result");
     gint64 status_code = _snapd_json_get_int (root, "status-code", 0);
     const gchar *kind = result != NULL ? _snapd_json_get_string (result, "kind", NULL) : NULL;
     const gchar *message = result != NULL ? _snapd_json_get_string (result, "message", NULL) : NULL;
+
+    if (error_value != NULL) {
+        *error_value = result != NULL ? json_object_get_member (result, "value") : NULL;
+        json_node_ref (*error_value);
+    }
 
     if (g_strcmp0 (kind, "login-required") == 0)
         g_set_error_literal (error,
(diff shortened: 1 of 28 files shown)


snapd-glib/requests/snapd-json.c
```diff
@@ -176,13 +176,18 @@ _snapd_json_get_date_time (JsonObject *object, const gchar *name)
...
+parse_error_response (JsonObject *root, JsonNode **error_value, GError **error)
 {
     JsonObject *result = _snapd_json_get_object (root, "result");
     gint64 status_code = _snapd_json_get_int (root, "status-code", 0);
     const gchar *kind = result != NULL ? _snapd_json_get_string (result, "kind", NULL) : NULL;
     const gchar *message = result != NULL ? _snapd_json_get_string (result, "message", NULL) : NULL;
 
+    if (error_value != NULL) {
+        *error_value = result != NULL ? json_object_get_member (result, "value") : NULL;
+        json_node_ref (*error_value);
```
robert-ancell, line 188: `json_node_ref` will hit an assertion if this in NULL (not documented, but I checked the json-glib source). So probably safer to do:
```c
JsonNode *node = NULL;
if (result != NULL)
    node = json_object_get_member (result, "value");
*error_value = node != NULL ? json_node_ref (node) : NULL;
```

jhenstridge: I've made the json_node_ref call conditional on *error_value being not-NULL.

robert-ancell:
Ah, the timeout out Qt test reminds me this is missing the Qt API for this - can you add it?

jhenstridge:
This PR doesn't change the public API of libsnapd-glib, so I'm not sure there is anything to change on the Qt side.

I tried restarting the test suite after seeing a similar failure on the Fedora run.  This time Fedora passed and ubuntu:rolling failed.  I wonder if this is just getting unlucky with slow IO? For example, I notice that this run went incredibly slowly for no obvious reason:

https://github.com/snapcore/snapd-glib/actions/runs/559596090

robert-ancell:
Oh, duh. This is just the first step. The other PR will have the Qt change. Sorry for the noise!

PR #96, opened 2021-02-09
ci: enble Github Actions based CI, using Travis script as a base

jhenstridge:
travis-ci.org has a banner saying it will be shut down in a few weeks, and that projects would be migrated to travis-ci.com.  It seems there is no longer an Open Source plan with the .com service, and the project would be switched to a trial plan that would eventually run out of minutes.

As an alternative, this PR ports the CI over to using Github Actions.  I've done a a mostly direct conversion of the Travis build script, only taking advantage of GH Actions built-in docker container syntax, and the conditional syntax for enabling container specific set up.

A sample run of the workflow can be seen here:

https://github.com/jhenstridge/snapd-glib/actions/runs/550765880

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,0 +1,43 @@
+name: test
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+
+jobs:
+  test:
+    strategy:
+      matrix:
+        image:
+          - ubuntu:rolling
+          - ubuntu:20.04
+          - ubuntu:18.04
+          - fedora:latest
+
+    runs-on: ubuntu-latest
+    container: ${{ matrix.image }}
+
+    steps:
+      - uses: actions/checkout@v2
+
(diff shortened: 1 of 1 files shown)


jhenstridge:
I also noticed that the travis script doesn't actually run the test suite: it looks like that was lost in the conversion from autotools, so I've added a "ninja test" step to the end of the workflow.

robert-ancell:
LGTM, can you make it delete `.travis.yml` at the same time?

robert-ancell:
Ah, just tried doing the same for LightDM. I see it won't actually do any of the actions until this lands on main. So a second MR to remove Travis probably makes more sense.

jhenstridge:
Yes.  `pull_request` workflows are only triggered if the target branch has a copy of the workflow (although it is the source branch's copy of the workflow that runs).

The example run I showed was triggered by a `push` event on my own branch before I restricted it to only check pushes to master.  You could use that trick to test workflows for other projects before merging.

PR #97, opened 2021-02-11
Add support for exit codes returned by snapctl commands

jhenstridge:
This builds on top of PR #95, which adds support for inspecting the value embedded in an error response.  The diff will be somewhat smaller once that is merged.

This PR adds new `snapd_client_run_snapctl2_*` functions that add a new `exit_code` out parameter.  These differ from the previous version in that if an error response is received with kind "unsuccessful", then the stdout, stderr, and exit code will be decoded from the response.  At present, `snapctl is-connected` is the only command that returns these responses, but more could in the future.

I've tried to keep the old `run_snapctl_*` functions the same, such that they will still set the GError if they receive an unsuccessful error.  Things would be a bit simpler if we didn't maintain that part of the API, but there would be no way for users of the old API to check error returns from is-connected otherwise.

I haven't updated the Qt binding yet, with similar concerns.  Switching QSnapdRunSnapCtlRequest to use the new API and adding an exitCode() method would be easy, but now code that was expecting a QSnapdError would no longer see it.

I do kind of wonder whether anyone has ever used this part of the snapd-glib API though, so it might not be so bad to take the simpler option.


snapd-glib/snapd-error.h
```diff
@@ -62,6 +62,7 @@ G_BEGIN_DECLS
  * @SNAPD_ERROR_NOT_A_SNAP: the given snap or directory does not look like a snap.
  * @SNAPD_ERROR_DNS_FAILURE: A hostname failed to resolve during the request.
  * @SNAPD_ERROR_OPTION_NOT_FOUND: A requested configuration option is not set.
+ * @SNAPD_ERROR_UNSUCCESSFUL: A snapctl command was unsuccessful.
```
robert-ancell, line 65: Does this error only apply to snapctl? The name of it suggests it could be due to anything.

jhenstridge: I copied the description from the corresponding constant in snapd:

https://github.com/snapcore/snapd/blob/9221806b95d60aab3e981b590c8440691ca1a898/client/errors.go#L124-L125

It is currently only used by the snapctl API, so it is accurate at present.

jhenstridge:
An alternative would be to continue returning a GError when we receive an "unsuccessful" response, but set stdout, stderr, and exit code.  This would allow the old API to be a simple wrapper around the new one (rather than having to generate an error where the new one doesn't), and do similar for the Qt API.

@robert-ancell: Do you have any preferences about this?

robert-ancell:
I'd prefer to only use a GError when the error is a true exception and no futher processing can be done. I think that better matches the way other GLib APIs use it. Please continue with the original proposal.

jhenstridge:
I was just looking through the snapd side for snapctl, and noticed that https://github.com/snapcore/snapd/pull/9692 had been merged.  If we're adding a new prototype for snapctl, then it should probably support stdin as an argument too so we don't need `run_snapctl3`.

jhenstridge:
After chatting with Michael, he suggests we ignore the current stdin support: it's only currently used by a full disk encryption setup command, and they don't like the current API because it requires draining stdin completely before making the REST API request.

Adding a `const char *stdin` arg to run_snapctl2 would just lock us in to a feature that snapd-glib users are unlikely to use and, and may not be compatible with whatever solution they come up with to replace the current naive version.

examples/snapctl-glib.c
```diff
@@ -0,0 +1,50 @@
...
+    args = g_new0 (char *, argc);
+    for (i = 0; i < argc-1; i++)
+        args[i] = argv[i+1];
+
+    if (!snapd_client_run_snapctl2_sync (client, context, args, &stdout_output, &stderr_output, &exit_code, NULL, &error)) {
+        fprintf(stderr, "error: %s\n", error->message);
+        return 1;
+    }
+    if (stdout_output) {
+        fputs(stdout_output, stdout);
```
robert-ancell, line 44: Missing whitespace after function name (and below).

snapd-qt/Snapd/client.h
```diff
@@ -873,6 +873,7 @@ class Q_DECL_EXPORT QSnapdRunSnapCtlRequest : public QSnapdRequest
     virtual void runAsync ();
     Q_INVOKABLE QString stdout () const;
     Q_INVOKABLE QString stderr () const;
+    Q_INVOKABLE int exit_code () const;
```
robert-ancell, line 876: Should be camel case - `exitCode`.

snapd-qt/client.cpp
```diff
@@ -3124,6 +3125,12 @@ QString QSnapdRunSnapCtlRequest::stderr () const
     return d->stderr_output;
 }
 
+int QSnapdRunSnapCtlRequest::exit_code () const
```
robert-ancell, line 3128: -> `exitCode`

robert-ancell:
Just a couple of small fixes, then LGTM. Thanks @jhenstridge!

PR #99, opened 2021-02-12
ci: collect Meson's log files on failed builds

jhenstridge:
Add an extra step to the CI to collect the meson log files if the job fails.  Each job has a unique artifact name so that logs from different matrix runs don't overwrite each other.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -22,6 +22,12 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - name: Initialise parameters
+        id: params
+        run: |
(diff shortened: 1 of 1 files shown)


robert-ancell:
Thanks @jhenstridge!

jhenstridge:
Okay.  It should actually work now.  The upload-artifact action did not like artifact names containing colons, so I needed to sanitise the image name before using it as an artifact name.

Here is a run where a job failed:

https://github.com/snapcore/snapd-glib/actions/runs/560008846

(I realise that it has used `logs-logs-*` as the name in this run: I've since corrected that).

PR #100, opened 2021-02-14
CI: Use Meson from the distribution rather than from PyPI

Conan-Kudo:
In practice, snapd-glib needs to work with the distribution-provided
Meson, so we should be testing against that. Since Fedora regularly
upgrades Meson, we should still get adequate testing as new versions
arrive.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -26,12 +26,12 @@ jobs:
         if: startsWith(matrix.image, 'ubuntu:')
         run: |
           apt-get update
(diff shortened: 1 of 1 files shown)


jhenstridge:
If you're opening/closing this to try and get data for every job in the matrix, try adding `fail-fast: false` to the strategy section of the job in `.github/workflows/test.yaml`.  That will prevent the other jobs from being cancelled on the first failure.

We'd discussed doing that to get similar behaviour to the old Travis job, but hadn't committed it yet.

Conan-Kudo:
@jhenstridge Ah, thanks for the tip! I'm new to GH Workflows, and wasn't sure how to get it to give me all the run output. 👍🏾

jhenstridge:
No problem.  The GH Actions workflow is less than a week old at the moment, so we're still working through the kinks (e.g. #99 was just merged to save log files in case of failure).

We'll probably get around to adding the fail-fast change elsewhere, but if you're working on some changes to the CI and it would help you finish those, you may as well include it here.

PR #108, opened 2021-10-12
Ensure async callbacks don't try to access destroyed Qt request objects.

robert-ancell:


aleixpol:
The change is rather verbose so I'm not sure I can fully see what it does, but making sure that there's something for the callbacks to call after destruction makes sense.

Thanks!

PR #109, opened 2021-10-19
Add support for snapd's desktop theme API

jhenstridge:
This PR adds support for snapd's desktop theme installation API, as found here:

https://github.com/snapcore/snapd/blob/master/daemon/api_themes.go

The API endpoint accepts GET and POST requests. The GET request is used to check the status of a list of GTK, icon, and sound themes, returning the status of each theme:
 * "installed" if an installed snap provides the theme via the content interface.
 * "available" if no installed snap provides the theme, but one exists on the store.
 * "unavailable" otherwise.

The POST request will start the install of the snaps associated with the list of theme names as an asynchronous task. It will error out if there is no work to do (i.e. all named themes are either installed or unavailable).

The GET request is open access, while the POST request requires authorisation (i.e. polkit in the case of desktop apps).

I haven't updated the Qt binding for this API yet: the PR was already quite large and I suspect there will only be a single user of the API, which will be using the glib version.

### DIFF
--- a/snapd-glib/snapd-client-sync.c
+++ b/snapd-glib/snapd-client-sync.c
@@ -1710,7 +1710,7 @@ snapd_client_download_sync (SnapdClient *self,
  * @cancellable: (allow-none): a #GCancellable or %NULL.
  * @error: (allow-none): #GError location to store the error occurring, or %NULL to ignore.
  *
- * Check whether the status of snap packaged versions of named desktop
+ * Check the status of snap packaged versions of named desktop
  * themes. For each theme, it will determine whether it is already
  * installed, uninstalled but available on the store, or unavailable.
  *
@@ -1739,9 +1739,9 @@ snapd_client_check_themes_sync (SnapdClient *self, GStrv gtk_theme_names, GStrv 
  * @cancellable: (allow-none): a #GCancellable or %NULL.
  * @error: (allow-none): #GError location to store the error occurring, or %NULL to ignore.
  *
- * Check whether the status of snap packaged versions of named desktop
- * themes. For each theme, it will determine whether it is already
- * installed, uninstalled but available on the store, or unavailable.
+ * Install snaps that provide the named desktop themes. If all the
+ * named themes are in the "installed" or "unavailable" states, then
+ * an error will be returned.
  *
  * Since: 1.60
  */


snapd-glib/snapd-client-sync.c
```diff
@@ -1697,3 +1697,62 @@ snapd_client_download_sync (SnapdClient *self,
...
+ * @gtk_theme_names: (allow-none): a list of GTK theme names.
+ * @icon_theme_names: (allow-none): a list of icon theme names.
+ * @sound_theme_names: (allow-none): a list of sound theme names.
+ * @gtk_theme_status: (out) (transfer container) (element-type utf8 SnapdThemeStatus): status of GTK themes.
+ * @icon_theme_status: (out) (transfer container) (element-type utf8 SnapdThemeStatus): status of icon themes.
+ * @sound_theme_status: (out) (transfer container) (element-type utf8 SnapdThemeStatus): status of sound themes.
+ * @cancellable: (allow-none): a #GCancellable or %NULL.
+ * @error: (allow-none): #GError location to store the error occurring, or %NULL to ignore.
+ *
+ * Check whether the status of snap packaged versions of named desktop
```
robert-ancell, line 1713: Should be "Check the status"?

snapd-glib/snapd-client-sync.c
```diff
@@ -1697,3 +1697,62 @@ snapd_client_download_sync (SnapdClient *self,
...
+ * @client: a #SnapdClient.
+ * @gtk_theme_names: (allow-none): a list of GTK theme names.
+ * @icon_theme_names: (allow-none): a list of icon theme names.
+ * @sound_theme_names: (allow-none): a list of sound theme names.
+ * @progress_callback: (allow-none) (scope call): function to callback with progress.
+ * @progress_callback_data: (closure): user data to pass to @progress_callback.
+ * @cancellable: (allow-none): a #GCancellable or %NULL.
+ * @error: (allow-none): #GError location to store the error occurring, or %NULL to ignore.
+ *
+ * Check whether the status of snap packaged versions of named desktop
```
robert-ancell, line 1742: Same comment as above.

robert-ancell (approved):
Apart from the small doc issue LGTM.

jhenstridge:
Thanks for the review.

I've updated the documentation. The body of the install_themes_sync() docs was incorrectly a copy of the check_themes_sync() docs, so I've changed that to something more appropriate.

PR #112, opened 2022-02-10
Add an example that reimplements the snap command line tool

robert-ancell:


robert-ancell:
I've added this example so we can easily check if the results from snapd-glib match what other snap tools.

PR #113, opened 2022-02-15
Add Qt implementation of themes API

robert-ancell:


snapd-qt/Snapd/client.h
```diff
@@ -899,6 +899,51 @@ class Q_DECL_EXPORT QSnapdDownloadRequest : public QSnapdRequest
...
+    Q_ENUM(ThemeStatus)
+
+    explicit QSnapdCheckThemesRequest (const QStringList& gtkThemeNames, const QStringList& iconThemeNames, const QStringList& soundThemeNames, void *snapd_client, QObject *parent = 0);
+    ~QSnapdCheckThemesRequest ();
+    virtual void runSync ();
+    virtual void runAsync ();
+    Q_INVOKABLE ThemeStatus gtkThemeStatus (const QString& name) const;
+    Q_INVOKABLE ThemeStatus iconThemeStatus (const QString& name) const;
+    Q_INVOKABLE ThemeStatus soundThemeStatus (const QString& name) const;
+    void handleResult (void *, void *);
```
mardy, line 923: You could avoid exposing this method, if you kept a pointer to `QSnapdCheckThemesRequest` in `QSnapdCheckThemesRequestPrivate`.

robert-ancell: Ah good point. All the existing classes follow this pattern, so I guess the damage is already done :(

PR #114, opened 2022-04-04
snapd-glib: use /v2/accessories/changes endpoint to check status of themes requests

jhenstridge:
This PR is intended to help fix https://github.com/snapcore/snapd-desktop-integration/issues/7, together with https://github.com/snapcore/snapd/pull/11450 on the snapd side. The snapd side has been merged and cherry-picked into the 2.55 branch.

The themes status API was intended to be usable from the restricted REST API available over `/run/snapd-snap.socket` to snaps that plug snap-themes-control. While the main requests function, we didn't expose a way to track the status of async tasks such as the one created by `snapd_client_install_themes_*`.

Rather than exposing the entire changes interface, a subset of it is available at `/v2/accessories/changes/NNN`, and will only provide details of changes initated through the themes interface.

This snapd-glib change makes SnapdRequestAsync responsible for creating the requests used to check the status of its change. A SnapdRequestAsync can then be marked as using the `/v2/accessories/changes` endpoint if appropriate. The SnapdPostThemes request does so.

As far as compatibility concerns go, this only changes the behaviour of the `snapd_client_install_themes_*` functions. They currently have only one user in the form of snapd-desktop-integration, which is broken without this change. So I think this should be a relatively safe change to merge.

### DIFF
--- a/snapd-glib/requests/snapd-get-change.c
+++ b/snapd-glib/requests/snapd-get-change.c
@@ -18,7 +18,7 @@ struct _SnapdGetChange
     gchar *change_id;
     SnapdChange *change;
     JsonNode *data;
-    gboolean accessories_change;
+    gchar *api_path;
 };
 
 G_DEFINE_TYPE (SnapdGetChange, snapd_get_change, snapd_request_get_type ())
@@ -55,9 +55,10 @@ _snapd_get_change_get_data (SnapdGetChange *self)
 }
 
 void
-_snapd_get_change_set_accessories_change (SnapdGetChange *self, gboolean accessories_change)
+_snapd_get_change_set_api_path (SnapdGetChange *self, const gchar *api_path)
 {
-    self->accessories_change = accessories_change;
+    g_free (self->api_path);
+    self->api_path = g_strdup (api_path);
 }
 
 static SoupMessage *
@@ -65,7 +66,7 @@ generate_get_change_request (SnapdRequest *request)
 {
     SnapdGetChange *self = SNAPD_GET_CHANGE (request);
 
-    g_autofree gchar *path = g_strdup_printf ("http://snapd%s/%s", self->accessories_change ? "/v2/accessories/changes" : "/v2/changes", self->change_id);
+    g_autofree gchar *path = g_strdup_printf ("http://snapd%s/%s", self->api_path ? self->api_path : "/v2/changes", self->change_id);
     return soup_message_new ("GET", path);
 }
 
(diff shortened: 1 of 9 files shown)


snapd-glib/requests/snapd-request-async.c
```diff
@@ -151,6 +173,9 @@ snapd_request_async_set_property (GObject *object, guint prop_id, const GValue *
     case PROP_PROGRESS_CALLBACK_DATA:
         priv->progress_callback_data = g_value_get_pointer (value);
         break;
+    case PROP_ACCESSORIES_CHANGE:
+        priv->accessories_change = g_value_get_boolean (value);;
```
robert-ancell, line 177: double semi-colon.

snapd-glib/requests/snapd-get-change.c
```diff
@@ -53,12 +54,18 @@ _snapd_get_change_get_data (SnapdGetChange *self)
     return self->data;
 }
 
+void
+_snapd_get_change_set_accessories_change (SnapdGetChange *self, gboolean accessories_change)
```
robert-ancell, line 58: I think this might be cleaner if instead `_snapd_get_change_set_path` is done which defaults to `/v2/changes` - that removes some of the special casing of what accessories are and allows another case in the future in case this is a pattern that snapd will continue to use.

jhenstridge: Okay. For what it is worth, `/v2/accessories/changes` was picked over `/v2/accessories/themes/changes` in the hope that we wouldn't add a third. It doesn't make much of a difference either way on the snapd-glib side though.

snapd-glib/requests/snapd-request-async.c
```diff
@@ -137,6 +139,26 @@ _snapd_request_async_report_progress (SnapdRequestAsync *self, SnapdClient *clie
...
+
+    _snapd_get_change_set_api_path (request, priv->change_api_path);
+    return request;
+}
+
+SnapdPostChange *
+_snapd_request_async_make_post_change_request (SnapdRequestAsync *self)
+{
+    SnapdRequestAsyncPrivate *priv = snapd_request_async_get_instance_private (self);
+    SnapdPostChange *request = _snapd_post_change_new (priv->change_id, "abort", NULL, NULL, NULL);
```
robert-ancell, line 156: I think the `"abort"` here should be a parameter as should the cancellable param etc. But in this case it's not a blocker as this is internal API.

robert-ancell (approved):
LGTM

PR #119, opened 2022-08-15
Don't treat unknown publisher validation as verified.

robert-ancell:


robert-ancell:
@mvo5 can you confirm if this is the correct behaviour? I believe at the time I was told that an unknown validation should be treated as verified, but this doesn't seem to be true anymore.

robert-ancell:
Going to assume this is correct, please should out if it's not...

PR #123, opened 2022-08-18
Build with soup3 by default, support soup2

robert-ancell:
Fixes https://github.com/snapcore/snapd-glib/issues/115

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -13,10 +13,16 @@ jobs:
       matrix:
         image:
           - ubuntu:rolling
-          - ubuntu:22.04
-          - ubuntu:20.04
-          - ubuntu:18.04
           - fedora:latest
+        soup:
+          - libsoup3
+        include:
+          - image: ubuntu:22.04
+            soup: libsoup2
+          - image: ubuntu:20.04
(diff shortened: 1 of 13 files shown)


robert-ancell:
@jbicha can you review this and see if I've got it suitable for packaging?

robert-ancell:
@aleixpol can you check if the Qt version numbers have been correctly updated? I've changed some of the directories to snapd-qt-2 and Snapd2, but I'm not sure if `/usr/include/snapd-qt-2/Snapd` should have a version number to match `/usr/lib/cmake/Snapd2`.

aleixpol:
The change looks good to me, CC to @jriddell who is more familiar with deb/ubuntu semantics than I am.

jbicha:
I haven't recompiled the reverse depends yet, but this merge proposal looks good enough to me.

I think bumping the API like this is useful. Thank you!

robert-ancell:
Thanks all! Let me know if there are any issues once released and we'll fix those ASAP.

robert-ancell:
Released in [snapd-glib 1.61](https://github.com/snapcore/snapd-glib/releases/tag/1.61)

PR #126, opened 2022-08-24
Fix library name in snapd-qt cmake config

robert-ancell:
Fixes https://github.com/snapcore/snapd-glib/issues/125


jbicha:
Robert, this fixed my build issue. Thanks!

PR #139, opened 2023-06-30
Add snapd_client_follow_changes

robert-ancell:
Once https://github.com/snapcore/snapd/pull/12898/ lands in snapd.


robert-ancell:
Upstream change didn't land - this is being designed differently.

PR #143, opened 2024-02-13
qt: Port to Qt 6

aleixpol:
It's what we are using in Discover now.


.github/workflows/test.yaml
```diff
@@ -55,7 +55,7 @@ jobs:
       - name: Install dependencies (Fedora)
         if: startsWith(matrix.image, 'fedora:')
         run: |
-          dnf install -y gcc gcc-c++ gobject-introspection-devel glib2-devel gtk-doc json-glib-devel libsoup3-devel ninja-build python3-pip qt5-qtbase-devel qt5-qtdeclarative-devel redhat-rpm-config vala
+          dnf install -y gcc gcc-c++ gobject-introspection-devel glib2-devel gtk-doc json-glib-devel libsoup3-devel ninja-build python3-pip qt6-qtbase-devel qt6-qtdeclarative-devel redhat-rpm-config vala
```
robert-ancell, line 58: Needs the equivalent for the ubuntu builds qtbase5-dev > qt6-base-dev, qtdeclarative5-dev > qt6-declarative-dev.

Note this means that qt support wont be available on 20.04 and 18.04 - so we need to decide if that's acceptable. I'm assuming we can't keep supporting qt5. The CI should disable qt builds on those older releases.

robert-ancell: The current goal is to be able to release from main to all supported Ubuntu releases. The Qt6 transition suggests we'll have to maintain a branch to keep supporting the older releases, might be worth making this snapd-glib 2.0. Note that 18.04 support ends in June 2023.

robert-ancell: If we do make this 2.0, then we can drop libsoup2 support at the same time.

aleixpol: By the look of the change, it should build on both Qt 5 and Qt 6. The only important part is that it can be built with Qt 6.

robert-ancell: Does it seem reasonable to have a build flag that chooses which one to build against? Or build both?

Conan-Kudo: I would say pick one.

robert-ancell:
Resolved in #147

PR #144, opened 2024-03-13
Add support for refresh-inhibited parameter

sergio-costas:
This PR implements in snapd-glib support for the 'refresh-inhibited' select parameter in the GET /v2/snap method, from the SD167 document ( https://docs.google.com/document/d/1HJQWKzgaB3yPKwRUqlxYhgyaKG5IcJe8eZOT4YY1CI8 )


snapd-glib/snapd-snap.c
```diff
@@ -424,6 +426,13 @@ snapd_snap_get_hold (SnapdSnap *self)
 }
 
 
+GDateTime *
+snapd_snap_get_proceed_time (SnapdSnap *self)
+{
+    g_return_val_if_fail (SNAPD_IS_SNAP (self), 0);
```
robert-ancell, line 432: 0 -> NULL

snapd-glib/snapd-snap.c
```diff
@@ -424,6 +426,13 @@ snapd_snap_get_hold (SnapdSnap *self)
 }
 
 
+GDateTime *
```
robert-ancell, line 429: Missing documentation

snapd-qt/Snapd/snap.h
```diff
@@ -86,6 +87,7 @@ class Q_DECL_EXPORT QSnapdSnap : public QSnapdWrappedObject
     bool devmode () const;
     qint64 downloadSize () const;
     QDateTime hold () const;
+    QDateTime proceed_time () const;
```
robert-ancell, line 90: proceed_time -> proceedTime

snapd-glib/snapd-snap.c
sergio-costas: Good catch! Fixed.

snapd-glib/snapd-snap.c
sergio-costas: Ops... you are right.

BTW: I see in all the documentation that it uses "@snap: a @SnapdSnap", but the parameter name in the function is `SnapdSnap *self`... Shouldn't it be "@self: a @SnapdSnap"?

snapd-qt/Snapd/snap.h
sergio-costas: fixed!

snapd-glib/snapd-snap.c
sergio-costas: Fixed.

robert-ancell: Responded in the other PR - the C files use 'self' and the headers/docs use the specific name.

snapd-glib/snapd-client.c
```diff
@@ -1951,13 +1951,17 @@ void
...
+                              GCancellable *cancellable,
+                              GAsyncReadyCallback callback,
+                              gpointer user_data)
 {
     g_return_if_fail (SNAPD_IS_CLIENT (self));
 
     g_autoptr(SnapdGetSnaps) request = _snapd_get_snaps_new (cancellable, names, callback, user_data);
     if ((flags & SNAPD_GET_SNAPS_FLAGS_INCLUDE_INACTIVE) != 0)
         _snapd_get_snaps_set_select (request, "all");
+    if ((flags & SNAPD_GET_SNAPS_FLAGS_REFRESH_INHIBITED) != 0)
```
robert-ancell, line 1963: These two flags are now mutually exclusive - should assert they're both not set.

snapd-glib/snapd-client-sync.c
```diff
@@ -512,6 +512,10 @@ G_GNUC_END_IGNORE_DEPRECATIONS
  * If @flags contains %SNAPD_GET_SNAPS_FLAGS_INCLUDE_INACTIVE then also return snaps
  * with status %SNAPD_SNAP_STATUS_INSTALLED.
  *
+ * If @flags contains %SNAPD_GET_SNAPS_FLAGS_REFRESH_INHIBITED, then it will return
```
robert-ancell, line 515: Document that only one of SNAPD_GET_SNAPS_FLAGS_INCLUDE_INACTIVE or SNAPD_GET_SNAPS_FLAGS_REFRESH_INHIBITED can be set.

snapd-glib/snapd-snap.c
```diff
@@ -423,6 +425,24 @@ snapd_snap_get_hold (SnapdSnap *self)
     return self->hold;
 }
 
+/**
+ * snapd_snap_get_proceed_time:
+ * @self: a @SnapdSnap
```
robert-ancell, line 430: @self->@snap

snapd-glib/snapd-client.c
sergio-costas: Added a check.

snapd-glib/snapd-client-sync.c
sergio-costas: Done.

sergio-costas: Mmm... but should I add a GError parameter, and set it if both flags are set?

snapd-glib/snapd-snap.c
sergio-costas: Yes, now I understand it.

sergio-costas: Done.

snapd-glib/snapd-client-sync.c
robert-ancell: I think an assert would be more appropriate - treat it as a programmer error to set both flags.

sergio-costas: You are right. Done.

robert-ancell:
@sergio-costas is this ready to land?

robert-ancell (approved):

PR #145, opened 2024-03-14
Fix gobject-introspection parameter name

sergio-costas:
The gobject-introspection documentation have incorrect parameter names for the "self" parameter. Probably, in origin specific names were used, and after that they were changed to "self", but the documentation comments weren't updated.

This patch fixes it.


robert-ancell:
It's a bit confusing but the header files use the specific names which matches the convention in other GObject libraries and the c files use self (which is clearer than the specific names). So it's currently correct.

PR #147, opened 2024-03-16
allow dual build of qt5 and qt6

carlosdem:
this work is based on apol's pull request https://github.com/snapcore/snapd-glib/pull/143.  it tries to support both qt5 and qt6 builds through some meson trickery inspired by the latest libappstream release.


Conan-Kudo:
This is a bad idea and in general probably unnecessary.

robert-ancell:
The value in this is we can ship these updated versions to older releases which use the old version of Qt5 version of Plasma discover. If we drop Qt5 support completely then we'd have to start a branch.

Conan-Kudo:
But do we need to be able to build both at once?

robert-ancell:
Right, I don't think there's any need to be parallel installed.

soumyaDghosh:
Hi, @carlosdem Can some ETA be given on this, considering, we'll be using this library to upgrade the discover backend. Please let me know, if I can help somehow.

carlosdem:
I'll finish it tomorrow (AEST time) and make it single build qt5 or qt6 only as requested.

soumyaDghosh:
> I'll finish it tomorrow (AEST time) and make it single build qt5 or qt6 only as requested.

Thanks a lot!

carlosdem:
> > I'll finish it tomorrow (AEST time) and make it single build qt5 or qt6 only as requested.
> 
> Thanks a lot!

@soumyaDghosh sorry life got in the way a bit.  new version is up.

robert-ancell:
Thanks @carlosdem! I've updated the CI so it now builds on the older releases.

PR #149, opened 2024-04-10
Allow to disable tests and examples build

sergio-costas:
This PR allows to disable the tests and the examples during build.

### DIFF
--- a/examples/meson.build
+++ b/examples/meson.build
@@ -1,9 +1,7 @@
-if get_option ('examples')
-  executable ('snap-glib',
-              'snap-glib.c',
-              dependencies: [snapd_glib_dep, gio_dep])
+executable ('snap-glib',
+            'snap-glib.c',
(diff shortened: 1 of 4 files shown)


examples/meson.build
```diff
@@ -1,7 +1,9 @@
-executable ('snap-glib',
-            'snap-glib.c',
-            dependencies: [snapd_glib_dep, gio_dep])
+if get_option ('examples')
```
robert-ancell, line 1: Would this work if the if was moved to the toplevel meson.build, i.e.:
```
if get_option('tests')
  subdir ('tests')
fi
if get_option('examples')
  subdir ('examples')
fi
```

sergio-costas: You are right, is much simpler that way. Done!

sergio-costas:
@robert-ancell This should be ready for merging.

robert-ancell (approved):

PR #150, opened 2024-04-12
Add nanosecond accuracy support

sergio-costas:
To work as expected, the NOTICES snapd API requires nanosecond accuracy. Unfortunately, curently GDateTime has only microsecond accuracy (and QDateTime has only millisecond accuracy), which means that some kind of trick is required to be able to pass nanosecond accuracy.

This patch implements that trick, by adding a method that allows to get the date/time value of the last occurred date in string format, as sent by snapd, and use it in the calls for reading more notices, in the AFTER field.


sergio-costas:
I also was thinking on creating a patch for GLib to add nanosecond accuracy to GDateTime. What do you think?

sergio-costas:
@robert-ancell Can you review this when you have some spare time? I also was thinking about creating a patch for GLib to add nanosecond accuracy to GDateTime. What do you think?

sergio-costas:
(about the patch for GDateTime, I think that is better to use the last changes in this PR).

snapd-glib/requests/snapd-json.c
```diff
@@ -300,6 +302,8 @@ _snapd_json_get_date_time (JsonObject *object, const gchar *name)
     if (timezone == NULL)
         timezone = g_time_zone_new_local ();
 
+    if (string_format != NULL)
+        *string_format = g_strdup(value);
```
robert-ancell, line 306: Why is this being returned as a string - isn't the remaining text the number of nanoseconds that we can return in a guint64 or similar?

sergio-costas: The idea is to put the exact same string that the server sent to us, with the whole date and time, when asking for the next notice. Also, the nanoseconds are passed in the string as a decimal value after the seconds, and the trailing zeros are removed, so it would require some pre- and post-processing, both for adding the trailing zeros when storing it as nanoseconds in the guint64, and also to take into account zeros immediately after the dot when converting it again to the string sent to the server. I already tried using a gdouble, but sometimes there is rounding errors in the conversion, resulting in the server sending the last notice over and over again. Also, this is only for "internal consumption": it is only to be used when asking for new notices, to specify the date of the last notice received, so it makes sense to just "paste it as-is" in the request message.

robert-ancell: Not processing the value has a bad code smell, while I understand in this case it's being used more as a token than a timestamp (also an indication of some bad architecture upstream) we should still be able to parse it. I don't think the overhead of the pre- and post-processing is significant in this case to warrant a workaround.

You should definitely not use a gdouble as it's a decimal fraction, but store it in an unsigned integer.

robert-ancell:
> I also was thinking on creating a patch for GLib to add nanosecond accuracy to GDateTime. What do you think?

This would be definitely be useful, worth opening a discussion with upstream.

snapd-glib/requests/snapd-json.c
sergio-costas: Done.

snapd-glib/snapd-client.c
```diff
@@ -1459,6 +1475,60 @@ snapd_client_get_notices_with_filters_finish (SnapdClient *self, GAsyncResult *r
...
+ *
+ * Since: 1.66
+ */
+void
+snapd_client_notices_set_after_notice (SnapdClient *self, SnapdNotice *notice)
+{
+    g_return_if_fail (SNAPD_IS_CLIENT (self));
+    SnapdClientPrivate *priv = snapd_client_get_instance_private (self);
+
+    priv->since_date_time_nanoseconds = (notice == NULL) ? G_MAXUINT : snapd_notice_get_last_occurred_nanoseconds (notice);
```
robert-ancell, line 1502: G_MAXUINT -> -1

snapd-glib/snapd-notice.c
```diff
@@ -184,6 +186,26 @@ snapd_notice_get_last_occurred (SnapdNotice *self)
...
 }
 
+/**
+ * snapd_notice_get_last_occurred_nanoseconds:
+ * @notice: a #SnapdNotice.
+ *
+ * Get the nanoseconds value of *last-occurred*, exactly as sent by
+ * snapd. Useful combined with #snapd_client_notices_set_since_nanoseconds, and used
+ * internally by snapd_client_notices_set_after_notice, to ensure the maximum
+ * possible precission when dealing with timestamps.
```
robert-ancell, line 196: Need to document it's -1 when unset.

robert-ancell (approved):
Looks good with the small fixes noted.

robert-ancell:
It seemed a little odd that the nanoseconds default to -1 instead of 0, but I think that's because you need to decide if it need to be used when generating the string, right?

snapd-glib/snapd-notice.c
sergio-costas: Ops... you are right.

sergio-costas: Done.

snapd-glib/snapd-client.c
sergio-costas: You are right. Fixed.

sergio-costas:
Ok, I added several comments and fixed what you asked for. Should be ready for merging (fingers crossed)

robert-ancell (approved):

PR #151, opened 2024-04-12
Object for monitoring new notices

sergio-costas:
This object allows to easily monitor new notices. It creates a new client connection (because the call is "blocking", in the sense that, when a client is asking for new notices in a socket, that socket can't accept new commands until a new notice is returned or the timeout has expired), and asks for new notices with a very big timeout. Whenever a new notice arrives, it sends a signal with that notice, and waits again for new notices.


sergio-costas:
I tested this object, compiling against #150, and works fine.

snapd-glib/snapd-notices-monitor.c
```diff
@@ -0,0 +1,166 @@
...
+        }
+    }
+    begin_monitor(self); // continue waiting for new notifications
+}
+
+static void begin_monitor(SnapdNoticesMonitor *self) {
+    snapd_client_set_notices_filter_by_date_seconds (self->client, self->last_date_time_seconds);
+    snapd_client_get_notices_async(self->client,
+                                   self->last_date_time,
+                                   200000000000000, // "infinity"
```
robert-ancell, line 80: G_INT64_MAX?

snapd-glib/snapd-notices-monitor.c
```diff
@@ -0,0 +1,166 @@
...
+
+    if (g_cancellable_is_cancelled(self->cancellable)) {
+        g_clear_object(&self->cancellable);
+        g_clear_pointer(&self->last_date_time, g_date_time_unref);
+        self->last_date_time_seconds = -1.0;
+        return;
+    }
+
+    if ((error == NULL) && (notices != NULL)) {
+        for (int i=0; i < notices->len; i++) {
```
robert-ancell, line 58: Whitespace around '='

snapd-glib/snapd-notices-monitor.c
```diff
@@ -0,0 +1,166 @@
...
+            g_autoptr(SnapdNotice) notice = g_object_ref(notices->pdata[i]);
+            g_autoptr(GDateTime) last_occurred = (GDateTime *)snapd_notice_get_last_occurred(notice);
+            gdouble last_occurred_seconds = snapd_notice_get_last_occurred_seconds(notice);
+
+            if ((self->last_date_time == NULL) ||
```
robert-ancell, line 63: Don't need all these parenthesis

snapd-glib/snapd-notices-monitor.c
```diff
@@ -0,0 +1,166 @@
...
+                  G_TYPE_NONE,
+                  1,
+                  G_TYPE_ERROR);
+}
+
+SnapdNoticesMonitor *snapd_notices_monitor_new(void) {
+    SnapdNoticesMonitor *self = g_object_new(snapd_notices_monitor_get_type(), NULL);
+    self->last_date_time_seconds = -1.0;
+    return self;
+}
```
robert-ancell, line 166: No newline at end of file.

snapd-glib/snapd-notices-monitor.c
```diff
@@ -0,0 +1,166 @@
...
+                  NULL,
+                  NULL,
+                  G_TYPE_NONE,
+                  1,
+                  G_TYPE_ERROR);
+}
+
+SnapdNoticesMonitor *snapd_notices_monitor_new(void) {
+    SnapdNoticesMonitor *self = g_object_new(snapd_notices_monitor_get_type(), NULL);
+    self->last_date_time_seconds = -1.0;
```
robert-ancell, line 164: Move this to `_init()`

snapd-glib/snapd-notices-monitor.c
```diff
@@ -0,0 +1,166 @@
...
+ *
+ * Starts the asynchronous listening proccess, that will wait for new
+ * notices and emit a "notice-event" signal with the new notice as
+ * parameter.
+ *
+ * @return: TRUE if there was an error, FALSE if everything worked fine
+ * and the object is listening for events.
+ */
+gboolean snapd_notices_monitor_start(SnapdNoticesMonitor *self,
+                                 GCancellable *cancellable) {
```
robert-ancell, line 99: It's probably better to make the cancellable internal, and instead provide a `snapd_notices_monitor_stop` if this is required. The cancellable is important as it should be cancelled in dispose - this will stop the program crashing if SnapdNoticesMonitor is unreferenced.

snapd-glib/snapd-notices-monitor.c
```diff
@@ -0,0 +1,166 @@
...
+/**
+ * snapd_notices_monitor_start:
+ * @monitor: a #SnapdNoticesMonitor
+ * @cancellable: a #GCancellable
+ *
+ * Starts the asynchronous listening proccess, that will wait for new
+ * notices and emit a "notice-event" signal with the new notice as
+ * parameter.
+ *
+ * @return: TRUE if there was an error, FALSE if everything worked fine
```
robert-ancell, line 95: This logic is the reverse of the norm - should return FALSE for error. And pass in a GError if information can be given about why the error occurred.

snapd-glib/snapd-notices-monitor.c
```diff
@@ -0,0 +1,166 @@
...
+static void monitor_cb(SnapdClient* source,
+                       GAsyncResult* res,
+                       SnapdNoticesMonitor *self) {
+
+    g_autoptr(GPtrArray) notices = NULL;
+    g_autoptr(GError) error = NULL;
+
+    notices = snapd_client_get_notices_finish(source, res, &error);
+
+    if (error != NULL) {
```
robert-ancell, line 42: Once using an internal cancellable - If error is G_IO_ERROR_CANCELLED then should just immediately return as the monitor may no longer exist.

snapd-glib/snapd-notices-monitor.c
```diff
@@ -0,0 +1,166 @@
...
+ * snapd_notices_monitor_get_client:
+ * @monitor: a #SnapdNoticesMonitor
+ *
+ * Returns the internal #SnapdClient used by this object. Useful if some
+ * kind of extra initialization must be done before begin to read notices.
+ * This client must NOT be used for anything else, because the get_notices
+ * call will block the connection until a response is generated. This client
+ * MUST be used only by this object.
+ *
+ * @return: (transfer none): a #SnapdClient
```
robert-ancell, line 117: `@return` -> Returns

snapd-glib/snapd-notices-monitor.h
```diff
@@ -0,0 +1,30 @@
...
+
+G_DECLARE_FINAL_TYPE(SnapdNoticesMonitor, snapd_notices_monitor, SNAPD,
+                     NOTICES_MONITOR, GObject)
+
+SnapdNoticesMonitor  *snapd_notices_monitor_new               ();
+
+SnapdClient          *snapd_notices_monitor_get_client        (SnapdNoticesMonitor     *monitor);
+
+gboolean              snapd_notices_monitor_start             (SnapdNoticesMonitor     *monitor,
+                                                               GCancellable        *cancellable);
```
robert-ancell, line 30: Final newline

snapd-glib/snapd-notices-monitor.c
sergio-costas: Already tried with G_MAXINT64, but it seems that there is a time limit in snapd and returns `invalid timeout: time: invalid duration "9223372036854775807us"`. I did some tests, and the maximum value seems to be around 9223300000000000.

snapd-glib/snapd-notices-monitor.c
sergio-costas: Fixed.

snapd-glib/snapd-notices-monitor.c
sergio-costas: Fixed.

snapd-glib/snapd-notices-monitor.c
sergio-costas: Fixed. And added an extension to Code to make it automagic :-)

snapd-glib/snapd-notices-monitor.c
sergio-costas: Good point. Done.

snapd-glib/snapd-notices-monitor.c
sergio-costas: Mmm... good point!

snapd-glib/snapd-notices-monitor.c
sergio-costas: The only error is if `self` is not a valid SnapdNoticesMonitor...

snapd-glib/snapd-notices-monitor.c
sergio-costas: Fixed.

snapd-glib/snapd-notices-monitor.h
sergio-costas: Fixed.

snapd-glib/snapd-notices-monitor.c
```diff
@@ -0,0 +1,185 @@
...
+ *
+ * Starts the asynchronous listening proccess, that will wait for new
+ * notices and emit a "notice-event" signal with the new notice as
+ * parameter.
+ *
+ * @return: FALSE if there was an error, TRUE if everything worked fine
+ * and the object is listening for events.
+ */
+gboolean snapd_notices_monitor_start(SnapdNoticesMonitor *self, GError **error) {
+
```
robert-ancell, line 94: Whitespace

snapd-glib/snapd-notices-monitor.c
```diff
@@ -0,0 +1,185 @@
...
+    }
+    g_cancellable_cancel(self->cancellable);
+    return TRUE;
+}
+
+/**
+ * snapd_notices_monitor_get_client:
+ * @monitor: a #SnapdNoticesMonitor
+ *
+ * Returns the internal #SnapdClient used by this object. Useful if some
```
robert-ancell, line 130: Is this necessary? If additional initialization is required perhaps there should be methods on this object to do those, or the client should be passed into `snapd_notices_monitor_new_with_client()`. Having a getter like this is a risk that a coding mistake could break things.

snapd-glib/snapd-notices-monitor.c
```diff
@@ -0,0 +1,185 @@
...
+ */
+SnapdClient *snapd_notices_monitor_get_client(SnapdNoticesMonitor *self) {
+    g_return_val_if_fail(SNAPD_IS_NOTICES_MONITOR(self), NULL);
+    return self->client;
+}
+
+static void snapd_notices_monitor_dispose(GObject *object) {
+    SnapdNoticesMonitor *self = SNAPD_NOTICES_MONITOR(object);
+
+    g_clear_object(&self->client);
```
robert-ancell, line 146: You should cancel the cancellable here, in case the object was unrefereced while still running.

snapd-glib/snapd-notices-monitor.c
```diff
@@ -0,0 +1,185 @@
...
+
+static void monitor_cb(SnapdClient* source,
+                       GAsyncResult* res,
+                       SnapdNoticesMonitor *self) {
+
+    g_autoptr(GError) error = NULL;
+    g_autoptr(GPtrArray) notices = snapd_client_get_notices_finish(source, res, &error);
+
+    if (g_cancellable_is_cancelled(self->cancellable)) {
+        g_clear_object(&self->cancellable);
```
robert-ancell, line 41: These clears seem unnecessary (and the ones below) - just the ones in the dispose should be fine.

snapd-glib/snapd-notices-monitor.c
```diff
@@ -0,0 +1,185 @@
...
+static void begin_monitor(SnapdNoticesMonitor *self);
+
+static void monitor_cb(SnapdClient* source,
+                       GAsyncResult* res,
+                       SnapdNoticesMonitor *self) {
+
+    g_autoptr(GError) error = NULL;
+    g_autoptr(GPtrArray) notices = snapd_client_get_notices_finish(source, res, &error);
+
+    if (g_cancellable_is_cancelled(self->cancellable)) {
```
robert-ancell, line 40: You should check the error is G_IO_ERROR_CANCELLED - because self may not be valid at this point (the object was unreferenced).

snapd-glib/snapd-notices-monitor.c
```diff
@@ -0,0 +1,185 @@
...
+ */
+gboolean snapd_notices_monitor_start(SnapdNoticesMonitor *self, GError **error) {
+
+    g_return_val_if_fail((error == NULL) || (*error == NULL), FALSE);
+    g_return_val_if_fail(SNAPD_IS_NOTICES_MONITOR(self), FALSE);
+    if (self->cancellable != NULL) {
+        *error = g_error_new(SNAPD_ERROR, SNAPD_ERROR_ALREADY_RUNNING, "The notices monitor is already running.");
+        return FALSE;
+    }
+    self->cancellable = g_cancellable_new();
```
robert-ancell, line 101: I would put this in init - then it's guaranteed to be available for the whole lifetime of the object.

snapd-glib/snapd-notices-monitor.c
sergio-costas: It makes sense, but then there's the problem of what happens with the callback. Will it be called after `dispose()`? I added a check in the callback `monitor_cb()` just in case, but I'm afraid that there could be an use-after-free case, because I'm passing `self` as the gpointer parameter. Should I be worried about that?

snapd-glib/snapd-notices-monitor.c
sergio-costas: I use the existence of a `cancellable()` object to know if the monitor is working, so I prefer to leave that one. The other two... yes, they can be removed.

snapd-glib/snapd-notices-monitor.c
sergio-costas: Ok, you explained two comments after this one: use G_IO_ERROR_CANCELLED.

snapd-glib/snapd-notices-monitor.c
```diff
@@ -0,0 +1,185 @@
...
+
+/**
+ * snapd_notices_monitor_start:
+ * @monitor: a #SnapdNoticesMonitor
+ *
+ * Starts the asynchronous listening proccess, that will wait for new
+ * notices and emit a "notice-event" signal with the new notice as
+ * parameter.
+ *
+ * @return: FALSE if there was an error, TRUE if everything worked fine
```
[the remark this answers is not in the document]
sergio-costas: Done

snapd-glib/snapd-notices-monitor.c
sergio-costas: Fixed.

snapd-glib/snapd-notices-monitor.c
sergio-costas: Good point... removing.

snapd-glib/snapd-notices-monitor.c
sergio-costas: done.

snapd-glib/snapd-notices-monitor.c
sergio-costas: I think that it's better to keep it that way because the existence of the cancellable allows to know if the monitor is running or not.

sergio-costas:
@robert-ancell This is ready for reviewing...

snapd-glib/snapd-notices-monitor.c
```diff
@@ -0,0 +1,206 @@
...
+/**
+ * snapd_notices_monitor_stop:
+ * @monitor: a #SnapdNoticesMonitor
+ *
+ * Stops the asynchronous listening proccess started with #snapd_notices_monitor_start.
+ *
+ * Returns: FALSE if there was an error, TRUE if everything worked fine.
+ */
+gboolean snapd_notices_monitor_stop (SnapdNoticesMonitor *self, GError **error) {
+
```
robert-ancell, line 115: whitespace

sergio-costas: Ops... also fixed the style to make it consistent with the rest of the code.

snapd-glib/snapd-client.c
```diff
@@ -1646,7 +1646,8 @@ snapd_client_get_change_finish (SnapdClient *self, GAsyncResult *result, GError
 
     if (!_snapd_request_propagate_error (SNAPD_REQUEST (request), error))
         return NULL;
-    return g_object_ref (_snapd_get_change_get_change (request));
+    SnapdChange *change = _snapd_get_change_get_change (request);
+    return change == NULL ? NULL : g_object_ref (change);
```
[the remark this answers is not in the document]
sergio-costas: I just put it "just in case", I presume. Changing it.

robert-ancell (approved):

PR #153, opened 2024-05-02
Fix snapd_client_get_change_finish()

sergio-costas:
The function can return NULL if there is no change with that ID.

### DIFF
--- a/snapd-glib/snapd-client.c
+++ b/snapd-glib/snapd-client.c
@@ -1632,7 +1632,7 @@ snapd_client_get_change_async (SnapdClient *self,
  * Complete request started with snapd_client_get_change_async().
  * See snapd_client_get_change_sync() for more information.
  *
- * Returns: (transfer full): a #SnapdChange or %NULL if there is no change with that ID.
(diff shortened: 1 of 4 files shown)


sergio-costas:
The check was needed...

robert-ancell:
Requesting an invalid ID should return an error, e.g.:
```
$ curl --unix-socket /var/run/snapd.socket http://localhost/v2/changes/1
{"type":"error","status-code":404,"status":"Not Found","result":{"message":"cannot find change with id \"1\""}}
```

So the ref change is appropriate but the documentation not?

sergio-costas:
The point is... I receive notifications with an specific change ID, but then I ask for that ID and I receive NULL. And the way the requests are implemented, the error isn't propagated as-is, so I can only return NULL. It seems that the request doesn't generate an error that can be propagated... I'll check the source code to see if I can modify it.

sergio-costas:
@robert-ancell Ok, now the error is propagated.

robert-ancell (approved):
Thanks!

PR #157, opened 2024-06-13
Add clang format code and script

sergio-costas:
This PR adds a .clang-format file and a script for automatically format the source code. It also adds a CI test.


sergio-costas:
@robert-ancell This patch should give you a break :-) But still requires some revision for the .clang-format file.

sergio-costas:
Something that I've been unable to make is to avoid whitespaces after g_autoptr...

sergio-costas:
If you like this approach, I'll run the script over the source and include in this PR the formatted code, to make it pass the test.

robert-ancell:
Looks great! I appreciate we can't exactly match the style we want, but that's not the top priority. The style should probably be the same as in snapd-desktop-integration, but again, not the top priority.

sergio-costas:
If you prefer the style from snapd-desktop-integration, then I only have to remove the .clang-format file... It seems to use the clang-format default style.

robert-ancell:
Confirming I have seen this, just haven't had the time to double check it - being such a big change don't want to have to make this change more than once!

sergio-costas:
I fully agree.

sergio-costas:
@robert-ancell @jhenstridge I changed this to use the same format than `snapd-desktop-integration`. What do you think? (I still have the old clang-format file if you prefer the old format).

robert-ancell (approved):
It's big and scary, but let's just pull off the plaster and go for it. It will make our life easier ongoing :)

sergio-costas:
I'm remembering the scene in Indiana Jones and the Last Crusade, when he has to cross the invisible bridge :-D

PR #159, opened 2024-07-01
Remove unnecessary use of const

robert-ancell:


sergio-costas (approved):
You are right: the const in the integers make no sense, and the const in the GDateTime objects are incorrect because we are returning them already ref'ed.

PR #160, opened 2024-07-01
Use correct prototype for function without args

robert-ancell:
https://stackoverflow.com/questions/51032/is-there-a-difference-between-foovoid-and-foo-in-c-or-c


sergio-costas (approved):
Good point!

PR #166, opened 2024-08-23
Make notices parsing consistent with other JSON parsing code

robert-ancell:


robert-ancell:
cc @sergio-costas

sergio-costas (approved):
Looks sensible. Go ahead.

PR #167, opened 2024-08-23
Don't use gint32 for nanoseconds.

robert-ancell:
The property is an int, so this might break on systems where sizeof(gint32) != sizeof(int) and may be the cause for https://bugs.launchpad.net/ubuntu/+source/snapd-glib/+bug/2077200

### DIFF
--- a/snapd-glib/requests/snapd-json.c
+++ b/snapd-glib/requests/snapd-json.c
@@ -259,11 +259,11 @@ is_timezone_prefix (gchar c)
     return c == '+' || c == '-' || c == 'Z';
 }
 
-static gint32
+static int
 parse_nanoseconds(gchar *token)
 {
     gchar *nanostr;
-    gint32 nanosecs;
+    int nanosecs;
     size_t nanostr_len;
 
     // find the decimal dot
@@ -285,7 +285,7 @@ parse_nanoseconds(gchar *token)
 }
 
 GDateTime *
-_snapd_json_get_date_time (JsonObject *object, const gchar *name, gint32 *nanoseconds)
+_snapd_json_get_date_time (JsonObject *object, const gchar *name, int *nanoseconds)
 {
     const gchar *value = _snapd_json_get_string (object, name, NULL);
     if (value == NULL)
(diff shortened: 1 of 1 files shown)


robert-ancell:
cc @sergio-costas

sergio-costas:
Shouldn't it be better to set the property as an "int32"?

sergio-costas:
Also, in ARMHF int seems to also be 32 bits...

robert-ancell:
The potential issue is the property is defined as an integer. If last_occurred_nanoseconds does not have the size expected, then all the following properties may be shifted. We could cast it to an integer at this point, but is there any reason not to store it as an integer everywhere?
```c
    g_autoptr(SnapdNotice) notice = g_object_new (SNAPD_TYPE_NOTICE,
                                                  "id", _snapd_json_get_string (object, "id", NULL),
                                                  "user-id", user_id,
                                                  "notice-type", notice_type,
                                                  "key", _snapd_json_get_string (object, "key", NULL),
                                                  "first-occurred", first_occurred,
                                                  "last-occurred", last_occurred,
                                                  "last-occurred-nanoseconds", last_occurred_nanoseconds,
                                                  "last-repeated", last_repeated,
                                                  "occurrences", _snapd_json_get_int (object, "occurrences", -1),
                                                  "expire-after", expire_after,
                                                  "repeat-after", repeat_after,
                                                  "last-data", last_data,
                                                  NULL);

```

And set as such. 
```c
    g_object_class_install_property (gobject_class,
                                     PROP_LAST_OCCURRED_NANO,
                                     g_param_spec_int ("last-occurred-nanoseconds",
                                                       "last-occurred-nanoseconds",
                                                       "Time this notice last ocurred, in string format and with nanosecond accuracy",
                                                       -1, 999999999, -1,
                                                       G_PARAM_READWRITE | G_PARAM_CONSTRUCT_ONLY));
```

robert-ancell:
This is a common bug I've seen with GObject properties, and is a huge problem as there's really no way to automatically pick this up. `g_object_new` is really unsafe. :(

sergio-costas:
Oh!!!! NOW I get it. The problem is in g_object_new()... so, the problem is that an int32 is passed to the function, but if internally it expects a different size, since there is no type checking because it is "dynamic", the compiler can't adjust the size, so the setter will take an incorrect value. If it expects a 64-bit int, it will add garbage because the setter will get 64 bits from the stack, thus the 32 bits of the value, and the next 32 bits from the next property name. And also the next property name will be "cut" by those four bytes, resulting in an "unknown property" error...

Good point, indeed...

sergio-costas (approved):
Absolutely right. Fixes a subtle but critical error. Thanks!

robert-ancell:
I think if you could make a tool to automatically check this then that would be great!

PR #169, opened 2024-08-27
Replace gint32 with int for nanoseconds

robert-ancell:
There's no need to limit it to 32 bits, and int will always be >= 32bit.


robert-ancell:
Proposing this for consistency, as we don't want the int -> int32 conversion that occurs to potentially change a value. Easier on the API user as they don't need to use a specific data type. Should otherwise not have any behaviour change.

sergio-costas (approved):
Looks sensible, and, as you said, avoids having to use specific types, which always is good.

PR #171, opened 2024-08-28
Fix test for 18.04

### LINKED ISSUE #168: 18.04 CI tests failing
Node seems to be broken?
```
/__e/node20/bin/node: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by /__e/node20/bin/node)
```

sergio-costas:
The tests in Bionic fail due to a problem in nodejs. Found a workaround at the end of https://github.com/actions/checkout/issues/1590 It consists on disable using node20 and allow using an insecure version of nodejs.

Fix #168

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -33,7 +33,17 @@ jobs:
     container: ${{ matrix.image }}
 
     steps:
-      - uses: actions/checkout@v2
+      - name: Setup environment
+        # Ubuntu 18.04 has a GLib version too old for nodejs20 or later,
+        # so, in order to do the tests, we must allow to use an older version.
+        # https://github.com/actions/checkout/issues/1590
+        # In that thread, Flamefire proposed this same solution:
+        # https://github.com/boostorg/nowide/commit/d4aa719f21149de8960247403435e9b794a01255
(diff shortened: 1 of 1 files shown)


sergio-costas:
@robert-ancell Please, review it to ensure that I did correctly. The workaround should be used only in 18.04.

.github/workflows/test.yaml
```diff
@@ -33,7 +33,12 @@ jobs:
     container: ${{ matrix.image }}
 
     steps:
-      - uses: actions/checkout@v2
+      - name: Setup environment
```
robert-ancell, line 36: Can you add a comment here why this occurs and linking to https://github.com/canonical/snapd-glib/issues/168. Hopefully we can remove this at some point.

robert-ancell (changes requested):
Looks good! I don't understand the workaround but as long as we leave a comment / link as to why this is here should be good.

.github/workflows/test.yaml
sergio-costas: AFAIU, nodejs 20 requires a newer GLib version than the one available in 18.04, so this patch specifies to not use nodejs20 and allow to use an insecure, old version.

PR #172, opened 2024-08-28
Add snapd-notice2 object

### LINKED ISSUE #170: SnapdNotice getters return references
The getters in SnapdNotice return references (e.g. `snapd_notice_get_last_repeated`). This isn't necessary (the object is not created and is owned by the object) and doesn't match the style of the other objects (and GObject code in general). This is a risk on the API user who might forget to release the reference and requires more boilerplate code for simple cases.

It would be nice if we correct this, I see two options:
- Make new getters and deprecate the old ones, e.g. `snapd_notice_get_last_repeated2`. It's a little ugly but works.
- Make a new object SnapdNotice2 that has new methods. More work than the above and even more messy.

sergio-costas:
Several getters in snapd-notice were returning references instead of pointers to internal objects. This patch creates the snapd-notice2 object which fixes this.

Fix #170


sergio-costas:
Maybe it's more clear to just add the new getters, and use `rename to ...` to ensure that bindings use the right method... What do you think?

sergio-costas:
BTW: this patch includes #169

robert-ancell:
> Maybe it's more clear to just add the new getters, and use `rename to ...` to ensure that bindings use the right method... What do you think?

Yes, I think the new getters is simpler, as this only changes a subset of the functions where renaming the class renames all the functions.

robert-ancell:
Obsoleted by https://github.com/canonical/snapd-glib/pull/173

PR #173, opened 2024-08-29
Fix notices2 getters

### LINKED ISSUE #170: SnapdNotice getters return references

sergio-costas:
Several getters were returning a reference of the data instead of just a pointer; this is, they incremented the refcount of the returned element. This isn't what the rest of the API does, so for uniformity, this PR adds new versions of the getters and deprecates the old ones.
    
Fix #170

### DIFF
--- a/snapd-glib/snapd-notice.c
+++ b/snapd-glib/snapd-notice.c
@@ -153,7 +153,7 @@ snapd_notice_get_last_data (SnapdNotice *self)
 }
 
 /**
- * snapd_notice_get_last_data2:
+ * snapd_notice_get_last_data2: (rename to snapd_notice_get_last_data)
  * @notice: a #SnapdNotice.
  *
  * Get the data of the notice.
@@ -190,7 +190,7 @@ snapd_notice_get_first_occurred (SnapdNotice *self)
 }
 
 /**
- * snapd_notice_get_first_occurred2:
+ * snapd_notice_get_first_occurred2: (rename to snapd_notice_get_first_occurred)
  * @notice: a #SnapdNotice.
  *
  * Get the time this notification first occurred.
@@ -229,7 +229,7 @@ snapd_notice_get_last_occurred (SnapdNotice *self)
 }
 
 /**
- * snapd_notice_get_last_occurred2:
+ * snapd_notice_get_last_occurred2: (rename to snapd_notice_get_last_occurred)
  * @notice: a #SnapdNotice.
  *
  * Get the time this notification last occurred.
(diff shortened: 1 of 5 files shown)


snapd-glib/snapd-notice.c
```diff
@@ -143,6 +143,7 @@ snapd_notice_get_key (SnapdNotice *self)
  * Returns: (transfer container): a HashTable with the data elements.
  *
  * Since: 1.65
+ * Deprecated: 1.66
```
robert-ancell, line 146: Should be:
```
Deprecated: 1.66: Use snapd_notice_get_last_data2()
```

snapd-glib/snapd-notice.c
```diff
@@ -170,6 +189,25 @@ snapd_notice_get_first_occurred (SnapdNotice *self)
...
+ * Returns: (transfer none): a #GDateTime.
+ *
+ * Since: 1.66
+ */
+GDateTime *
+snapd_notice_get_first_occurred2 (SnapdNotice *self)
+{
+    g_return_val_if_fail (SNAPD_IS_NOTICE (self), NULL);
+    if (self->first_occurred == NULL)
+        return NULL;
```
robert-ancell, line 207: This doesn't do anything - it would have returned NULL in the next line anyway!

snapd-qt/notice.cpp
```diff
@@ -60,17 +60,17 @@ static QDateTime convertDateTime (GDateTime *datetime)
 
 QDateTime QSnapdNotice::firstOccurred () const
 {
-    return convertDateTime ((GDateTime*) snapd_notice_get_first_occurred (SNAPD_NOTICE (wrapped_object)));
+    return convertDateTime ((GDateTime*) snapd_notice_get_first_occurred2 (SNAPD_NOTICE (wrapped_object)));
```
robert-ancell, line 63: These casts shouldn't be necessary, right?

robert-ancell (approved):
Other than the above comments looks good, thanks!

snapd-glib/snapd-notice.c
sergio-costas: Fixed!

snapd-glib/snapd-notice.c
sergio-costas: Ops!!!! Yes, it was the check to avoid errors when calling XXX_ref. The problems of copy&paste 😅

snapd-qt/notice.cpp
sergio-costas: Mmm... yes, they aren't needed anymore. I added there when the return value was `const GDateTime*`, so they were required, but now they aren't. Fixed.

robert-ancell (approved):

PR #176, opened 2024-09-10
Fix build in ARMHF

sergio-costas:
Since in 64bit architectures, long int is the same than long long int, using %ld or %lu is the same than %lld and %llu. But this isn't the case for 32bit architectures.


sergio-costas:
Closing because there's a more generic patch for master.

Issue #162, opened 2024-07-31
Tests fail on armhf - log provided


ScarlettGatelyMoore:
Please see https://launchpadlibrarian.net/739560793/buildlog_ubuntu-oracular-armhf.snapd-glib_1.65-0ubuntu0~ubuntu24.10~ppa3_BUILDING.txt.gz

rikmills:
For reference, we (Kubuntu/Lubuntu/UbuntuStudio) need to get the qt6 port v1.65 or greater into Ubuntu oracular before feature freeze. Needed for Qt6 plasma-discover software centre.

robert-ancell:
@sergio-costas can you have a look at this? I suspect it was added in cc743d3d but I couldn't see anything obvious that would break on armhf.

rikmills:
snapd-glib v1.65 is now in Ubuntu Oracular proposed, with build time tests skipped as we have the autopkgtest infra there. 

See: https://launchpad.net/ubuntu/+source/snapd-glib/1.65-0ubuntu1

As was predictable, we now see this test failure manifest itself in the autopkgtest results:

https://autopkgtest.ubuntu.com/packages/s/snapd-glib/oracular/armhf

Which results in the package being blocked from migration to the release pocket:

https://ubuntu-archive-team.ubuntu.com/proposed-migration/update_excuses.html#snapd-glib

So any help/solution would be great, as this is now a real blocker to getting dependant packages (plasma-discover etc) into oracular.

Thanks

sergio-costas:
I returned today from holidays. I'll check this ASAP.

ScarlettGatelyMoore:
Any luck on this? We had to disable snaps in Kubuntu software store Discover, which is not something I want.

rikmills:
@robert-ancell @sergio-costas as above

sergio-costas:
Sorry for the delay... I'll check it right now.

sergio-costas:
Ok, I suspect that the problem is either in mock-snapd.c:

![imagen](https://github.com/user-attachments/assets/c65878f3-dc95-4bc2-b70c-b67b68c6cc01)

or in snapd-json.c:

![imagen](https://github.com/user-attachments/assets/923b7639-c19c-4650-904f-242b6c385547)

Maybe in armhf the type is neither G_TYPE_INT64 nor STRING, and it returns NULL?

rikmills:
> @ScarlettGatelyMoore @rikmills Can you test #176 please?

[...]

PR #177, opened 2024-09-10
Fix build in 32bit systems

sergio-costas:
At least in ARMhf systems, the current code fails due to the difference in sizes for long int and long long int. This patch fixes it.

Fix https://github.com/canonical/snapd-glib/issues/162

### DIFF
--- a/snapd-glib/requests/snapd-get-notices.c
+++ b/snapd-glib/requests/snapd-get-notices.c
@@ -102,8 +102,8 @@ generate_get_snap_request (SnapdRequest *request, GBytes **body)
     }
     if (self->timeout != 0) {
         add_uri_parameter_base (query, "timeout");
-        // timeout in microseconds
-        g_string_append_printf (query, "%lluus", (unsigned long long int)self->timeout);
+        // timeout in microseconds. self->timeout is a GTimeSpan, which is a typedef into a gint64 type
+        g_string_append_printf (query, "%" G_GINT64_FORMAT "us", self->timeout);
     }
 
     if (query->len != 0)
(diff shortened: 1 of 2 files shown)


snapd-glib/requests/snapd-get-notices.c
```diff
@@ -103,7 +103,7 @@ generate_get_snap_request (SnapdRequest *request, GBytes **body)
     if (self->timeout != 0) {
         add_uri_parameter_base (query, "timeout");
         // timeout in microseconds
-        g_string_append_printf (query, "%luus", self->timeout);
+        g_string_append_printf (query, "%lluus", (unsigned long long int)self->timeout);
```
robert-ancell, line 106: You should be able to do this: `"%" G_GINT64_FORMAT, self->timeout`.

robert-ancell (approved):
With the use of the define this should be good, thanks!

snapd-glib/requests/snapd-get-notices.c
jbicha: Should we use the approach from https://gitlab.gnome.org/GNOME/gnome-calendar/-/merge_requests/492 ?

robert-ancell: In this case timeout is a GTimeSpan, so it's never been explicitly chosen as an integer.

sergio-costas: Oh, clever!!! Thanks for the tip!

sergio-costas: Done!

PR #179, opened 2024-09-16
Generate documentation using gi-docgen

sergio-costas:


sergio-costas:
My idea is to also publish it directly in github every time master is updated. What do you think?

robert-ancell (approved):
Looks great! Updating the docs to the latest systems has been a long standing task I haven't investigated. Regarding where the generated docs go, I don't have a strong opinion other than we should do what the common pattern is here. Any existing libraries that we should use for reference?

Good to land when the CI is fixed.

sergio-costas:
Since gi-docgen is still experimental, I think it's better to stay with gtk-doc.

PR #180, opened 2024-09-17
Generate documentation on push to master

sergio-costas:
It seems that gi-docgen is still experimental, so I decided to maintain gtk-doc and just create a workflow to generate the HTML and upload it to the pages of snapd-glib.


robert-ancell (approved):
LGTM

PR #181, opened 2024-09-18
Add snapd_client_get_model

robert-ancell:
Fixes https://github.com/canonical/snapd-glib/issues/56


snapd-glib/snapd-client-sync.c
```diff
@@ -1964,3 +1964,53 @@ snapd_client_get_notices_with_filters_sync (SnapdClient *self,
...
+
+/**
+ * snapd_client_get_model_assertion_sync:
+ * @client: a #SnapdClient.
+ * @cancellable: (allow-none): a #GCancellable or %NULL.
+ * @error: (allow-none): #GError location to store the error occurring, or %NULL to ignore.
+ *
+ * Synchronously get the active model assertion for this device.
+ *
+ * Returns: an assertion or %NULL on error.
```
sergio-costas, line 1976: Shouldn't it include `(transfer full) (allow-none)`?

And the same in the other three methods (`get_model_assertion_finish`, `get_serial_assertion_sync` and `get_serial_assertion_finish`).

sergio-costas: Also, why is it an "assertion"? Is it because it contains JSON data instead of the serial value itself?

tests/test-qt.cpp
```diff
@@ -1068,7 +1068,6 @@ test_logout_sync ()
     QStringList discharges;
     for (int i = 0; d[i] != NULL; i++)
         discharges << d[i];
-    qDebug() << discharges;
```
sergio-costas, line 1071: Why did you remove this?

sergio-costas (changes requested):
Some nitpicks.

tests/test-qt.cpp
robert-ancell: This an unrelated fix, I really should have put in a separate PR - https://github.com/canonical/snapd-glib/pull/183

snapd-glib/snapd-client-sync.c
robert-ancell: I think it's not supposed to have `allow-none`, since the function only returns `NULL` on an error. In a nullable language the API user isn't supposed to handle the `NULL` case, this only occurs during an exception.

It's not "string" because the gchar * already implies it's a string. The string is a textual description of an assertion. I originally had it return `SnapdAssertion` but the existing assertion APIs use strings described like this, so I did it the same for consistency.

sergio-costas (approved):
Looks good.

PR #184, opened 2024-09-18
Update README to point to the docs

robert-ancell:


sergio-costas:
Looks good, but as I commented on mattermost, it isn't really updated because the contents file must be updated manually. I'll give another spin to gi-docgen, which does it fully automated.

PR #186, opened 2024-09-19
Use gi-docgen to generate documentation

sergio-costas:
This PR uses gi-docgen to generate the documentation.


robert-ancell (approved):
Awesome, thanks!

PR #188, opened 2024-09-23
Added missing Since fields

sergio-costas:
Some `Since` and other documentation fields were missing.

### DIFF
--- a/snapd-glib/snapd-notices-monitor.c
+++ b/snapd-glib/snapd-notices-monitor.c
@@ -221,11 +221,10 @@ SnapdNoticesMonitor *snapd_notices_monitor_new(void) {
 
 /**
  * snapd_notices_monitor_new_with_client:
+ * @client: a #SnapdClient object
  *
  * Creates a new #SnapdNoticesMonitor to receive events, using the
  * specified #SnapdClient object to ask info about each task.
- *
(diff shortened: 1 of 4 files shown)


snapd-glib/snapd-notices-monitor.c
```diff
@@ -17,6 +17,28 @@
...
+ * due to the snap being active, or inhibited launches due to an ongoin
+ * refresh.
+ */
+
+/**
+ * SnapdNoticesMonitor:
+ *
+ * #SnapdNoticesMonitor allows to receive in real time events from
+ * snapd, like status changes in an ongoin refresh, inhibited refreshes
+ * due to the snap being active, or inhibited launches due to an ongoin
```
robert-ancell, line 36: ongoin -> ongoing

robert-ancell: (twice in this paragraph)

robert-ancell (approved):
Small spelling mistake but otherwise good to go!

snapd-glib/snapd-notices-monitor.c
sergio-costas: Agh! The spell checker was disabled... Fixed. Sorry O:-)

PR #189, opened 2024-09-23
Don't use deprecated calls

sergio-costas:
When creating the new snapd_notice_get... calls and deprecating the old ones, this one was left for replacement.


robert-ancell (approved):
LGTM

PR #191, opened 2024-09-25
Fix GIR comments

sergio-costas:
The autoformatting has the problem of cutting comments in several lines, which is a problem for the GIR parser if a GIR line with parentheses is cut and the opening and closing parentheses are in different lines.

This change instructs clang-format to not reflow comments. It also fixes the GIR errors.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,1 @@
+ReflowComments: false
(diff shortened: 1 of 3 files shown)


sergio-costas:
@robert-ancell I'll merge this for now to ensure that the documentation is fine.

PR #194, opened 2024-11-04
Fix memory leaks in mock

sergio-costas:
The snapd mock had some leaks. These patches fix them, to allow to find future leaks more easily.

### DIFF
--- a/tests/mock-snapd.c
+++ b/tests/mock-snapd.c
@@ -4948,7 +4948,7 @@ static void handle_notices(MockSnapd *self, SoupServerMessage *message,
     }
     guint data_size = g_hash_table_size(notice->last_data);
     if (data_size != 0) {
-      gchar **keys =
+      g_autofree gchar **keys =
           (gchar **)g_hash_table_get_keys_as_array(notice->last_data, NULL);
       json_builder_set_member_name(builder, "last-data");
       json_builder_begin_object(builder);
@@ -4958,7 +4958,6 @@ static void handle_notices(MockSnapd *self, SoupServerMessage *message,
         json_builder_add_string_value(builder, value);
       }
       json_builder_end_object(builder);
-      g_free(keys);
     }
     json_builder_end_object(builder);
   }
--- a/tests/mock-snapd.c
+++ b/tests/mock-snapd.c
@@ -1692,8 +1692,8 @@ static void mock_change_free(MockChange *change) {
   g_clear_list(&change->tasks, (GDestroyNotify)mock_task_free);
 #else
   if (change->tasks) {
-    g_list_free_full(change->task, (GDestroyNotify)mock_task_free);
-    change->taks = NULL;
+    g_list_free_full(change->tasks, (GDestroyNotify)mock_task_free);
+    change->tasks = NULL;
   }
 #endif
   g_clear_pointer(&change->data, json_node_unref);
--- a/tests/mock-snapd.c
+++ b/tests/mock-snapd.c
@@ -1679,7 +1679,7 @@ static void mock_task_free(MockTask *task) {
     mock_snap_free(task->snap);
   g_free(task->snap_name);
   g_free(task->error);
-  g_ptr_array_free(task->affected_snaps, TRUE);
+  g_ptr_array_unref(task->affected_snaps);
   g_slice_free(MockTask, task);
 }
 
--- a/tests/mock-snapd.c
+++ b/tests/mock-snapd.c
@@ -1678,6 +1678,7 @@ static void mock_task_free(MockTask *task) {
     mock_snap_free(task->snap);
   g_free(task->snap_name);
   g_free(task->error);
+  g_ptr_array_free(task->affected_snaps, TRUE);
   g_slice_free(MockTask, task);
 }
 
--- a/tests/mock-snapd.c
+++ b/tests/mock-snapd.c
@@ -1675,11 +1675,10 @@ static void mock_task_free(MockTask *task) {
   g_free(task->progress_label);
   g_free(task->spawn_time);
   g_free(task->ready_time);
-  if (task->snap != NULL)
-    mock_snap_free(task->snap);
+  g_clear_pointer(&task->snap, mock_snap_free);
   g_free(task->snap_name);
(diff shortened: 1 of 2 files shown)


tests/mock-snapd.c
```diff
@@ -290,9 +290,9 @@ static void mock_notice_free(MockNotice *notice) {
   g_free(notice->type);
   g_free(notice->expire_after);
   g_free(notice->repeat_after);
-  g_date_time_unref(notice->first_occurred);
-  g_date_time_unref(notice->last_occurred);
-  g_date_time_unref(notice->last_repeated);
+  g_clear_pointer(&notice->first_occurred, g_date_time_unref);
+  g_clear_pointer(&notice->last_occurred, g_date_time_unref);
+  g_clear_pointer(&notice->last_repeated, g_date_time_unref);
```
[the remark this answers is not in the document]
sergio-costas: Not really: g_date_time_unref() dislikes being called with a NULL pointer, so adding g_clear_pointer really fixes some WARNING messages.

tests/mock-snapd.c
```diff
@@ -1678,6 +1679,7 @@ static void mock_task_free(MockTask *task) {
     mock_snap_free(task->snap);
   g_free(task->snap_name);
   g_free(task->error);
+  g_ptr_array_free(task->affected_snaps, TRUE);
```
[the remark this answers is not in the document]
sergio-costas: Done.

tests/mock-snapd.c
```diff
@@ -4956,6 +4958,7 @@ static void handle_notices(MockSnapd *self, SoupServerMessage *message,
         json_builder_add_string_value(builder, value);
       }
       json_builder_end_object(builder);
+      g_free(keys);
```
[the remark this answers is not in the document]
sergio-costas: Changed to g_autofree(). Also, about changing that... the code was already that way, so I prefer to touch it as less as possible.

tests/mock-snapd.c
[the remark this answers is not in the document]
sergio-costas: Also... what is a "sweated commit"? Google says nothing...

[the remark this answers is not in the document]
3v1n0: lol, sorry... I was using mobile on first comment and for some reason the auto-correction changed it to nonsense.

I meant, "separated commit".

tests/mock-snapd.c
```diff
@@ -1687,9 +1688,15 @@ static void mock_change_free(MockChange *change) {
   g_free(change->summary);
   g_free(change->spawn_time);
   g_free(change->ready_time);
-  g_list_free_full(change->tasks, (GDestroyNotify)mock_task_free);
-  if (change->data != NULL)
-    json_node_unref(change->data);
+#ifdef GLIB_VERSION_2_64
```
3v1n0, line 1691: If there are no SRU-concerns I think is totally fine to bump the glib dependency: 2.64 it's 4 years old already so I don't think we need to care about it too much.

robert-ancell: This was here so we could support 18.04 LTS, if that's no longer a concern then bumping it to 2.64 means the minimum supported version is 20.04 LTS (at least without patching to support 18.04).

tests/test-glib.c
```diff
@@ -824,7 +824,7 @@ static void test_logout_async(void) {
   snapd_client_set_auth_data(client, auth_data);
 
   g_assert_nonnull(mock_snapd_find_account_by_id(snapd, id));
-  AsyncData *data = async_data_new(loop, snapd);
+  g_autoptr(AsyncData) data = async_data_new(loop, snapd);
```
robert-ancell, line 827: This is a leak - the data is passed into the callback. (Other cases below).

robert-ancell (changes requested):
This has added some leaks.

tests/test-glib.c
sergio-costas: Yes, it is passed, but then we wait in a local mainloop(), so the data struct is freed after the g_mainloop_quit() called in the callback, when the control returns to the function that created the memory block. I think that this is more clear that having the creation in a function and the destruction in another one. So really there is no leak (in fact, there were two leaks before this PR because I created the memory block but forgot to free it in the callback; that's why I changed everything to g_autoptr).

tests/mock-snapd.c
sergio-costas: You are right. Removed the #ifdefs and left the code as before.

tests/mock-snapd.c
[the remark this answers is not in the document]
sergio-costas: I'll put in a separate PR.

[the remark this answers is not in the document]
sergio-costas: Ok, now the tests fail due to NULL being passed to g_date_time_unref() 😅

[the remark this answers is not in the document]
sergio-costas: https://github.com/canonical/snapd-glib/pull/196 fixes it.

tests/test-glib.c
sergio-costas: The main problem is that, sometimes, the callbacks are expected to be called several times, so it requires to think carefully when to free the passed data. Instead, using a g_autoptr() in the initial function, solves the problem automagically (as long as g_main_loop_run() is used, of course).

sergio-costas: Of course, the alternative is to remove the `g_autoptr()` that I added in the creation functions, and add an `async_data_free()` after each `g_main_loop_quit()` call in the callbacks, but in my opinion, that is more complex in this case.

robert-ancell: I understand now. I think we'd want to be careful doing it this way, as while it does make the code simpler, it's inconsistent with the other code.

robert-ancell (approved):
LGTM

tests/test-glib.c
sergio-costas: Mmm... I think that it is better to keep consistency... I'll revert this part.

sergio-costas: Ok, reverted and added some extra comments. It should be ready.

PR #196, opened 2024-11-05
Cleanup use of unref

sergio-costas:
g_XXXX_unref calls dislike being called with a NULL parameter, so this little cleanup ensures that they are called only when there is something to unref, thus removing warnings.


sergio-costas:
It's not a warning, but a critical that exits the program, so it is really required to check for NULL before calling the free function.

About using an `if`... it is exactly what `g_clear_pointer()` does under the hood; I just think it's more elegant to use it instead of the `if`...

nteodosio (approved):

PR #197, opened 2024-11-11
Added test support for "after"

sergio-costas:
The mock-snapd class allows to simulate receiving notices, but until now it didn't honor the "after" field, so it always returned all the notices.

This patch changes this, also allowing to test that the "after" parameter is correctly parsed. This, also, allowed to find a bug when requesting notices after a specific date when using GLib 2.65 or previous.

### DIFF
--- a/tests/mock-snapd.c
+++ b/tests/mock-snapd.c
@@ -20,6 +20,53 @@
 
 #include "mock-snapd.h"
 
+#if !GLIB_CHECK_VERSION(2, 66, 0)
+
+#define HEX_CHAR_TO_NUMBER(c) (((c) < 'a') ? (c) - '0' : ((c) - 'a' + 10))
+
+// a quick&dirty replacement to allow the tests to work, since g_uri_parse_params
+// wasn't added until GLib 2.66
+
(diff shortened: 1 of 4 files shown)


tests/mock-snapd.c
```diff
@@ -20,6 +20,53 @@
...
 
+#if !GLIB_CHECK_VERSION(2, 66, 0)
+
+#define HEX_CHAR_TO_NUMBER(c) (((c) < 'a') ? (c) - '0' : ((c) - 'a' + 10))
+
+// a quick&dirty replacement to allow the tests to work, since g_uri_parse_params
+// wasn't added until GLib 2.66
+
+// Replaces in-place any %XY token with the corresponding 0xXY byte
+static void parse_string(gchar *str) {
```
robert-ancell, line 31: This could probably do with a better name.

robert-ancell (approved):
Other than the naming, looks good.

tests/mock-snapd.c
sergio-costas: You are right. Changed to `decode_url_escape_codes()`.

Issue #175, opened 2024-09-08
build fail on release 1.65 and 1.66


tristanRW:
When fetching the newest release 1.65 and running `meson setup -Dprefix=/usr _builddir_` followed by `ninja -C _builddir_` ninja fails at [147/205]: tests/tests-markdown-qt6.

I'm at my wits end with this, after researching and trying for days, but can provide any logs needed to someone skilled enough.

[meson.log](https://github.com/user-attachments/files/16956424/meson.log)
[ninja.log](https://github.com/user-attachments/files/16956425/ninja.log)

sergio-costas:
Try to delete the _builddir_ folder before running setup. The log says that it's already configured, so it maybe is using an old setup.

tristanRW:
Please refer to the ninja.log. The builddir was configured using the command given. I will attach the correct meson.log soon.

sergio-costas:
The point is that I need info about which system type and architecture is this being compiled on, and that is shown in the meson setup log...

tristanRW:
I finally updated both logs. Should contain all info now. Sorry for the delay and thank you for looking into it so quickly.

sergio-costas:
Mmm... Under which Ubuntu version are you building this?

sergio-costas:
OK, I tested it under Ubuntu Oracular, which has the same gcc version, and still builds... Can you tell me under which system are you building it, please?

tristanRW:
Hi, I am building on Arch x86_64, qt5 and qt6 are installed for plasma6-desktop.
Where should I check for the .so files (where should they be)?

sergio-costas:
Check where in your system are these files:

```
libQt6Qml.so
libQt6QmlBuiltins.a
libQt6Network.so
libQt6Core.so
```
If they are at /usr/lib, then check with `ldd` if all the dependencies are in your system.

tristanRW:
Sorry, I am new to this. Which file should I check using ldd ?

sergio-costas:
Each one of those four libraries, wherever they are. Let's say that they are at /usr/lib/x86_64-linux-gnu; then you should do

`ldd /usr/lib/x86_64-linux-gnu/libQt6Qml.so`

and check that all entries point to another file.

Although, if KDE works, that should not be a problem...

tristanRW:
Ok, so:

[...]

PR #199, opened 2025-02-24
Fix build when Qt is compiled with QT_USE_PROTECTED_VISIBILITY

sergio-costas:
When QT_USE_PROTECTED_VISIBILITY is used to compile Qt (like in Arch), any library compiled against it will, by default, do the same, so all their symbols will be protected by default too.

This patch fixes this, doing the same that CALAMARES did https://github.com/calamares/calamares/commit/4eba859236777c5764ebd3db883a303ed810ee7b

But maybe a different approach would be better...

Fix https://github.com/canonical/snapd-glib/issues/175

### DIFF
--- a/snapd-qt/Snapd/Category
+++ b/snapd-qt/Snapd/Category
@@ -13,7 +13,9 @@
 #include <QtCore/QObject>
 #include <Snapd/WrappedObject>
 
-class Q_DECL_EXPORT QSnapdCategory : public QSnapdWrappedObject
+#include "snapdqt_global.h"
+
+class LIBSNAPDQT_EXPORT QSnapdCategory : public QSnapdWrappedObject
 {
     Q_OBJECT
 
--- a/snapd-qt/Snapd/alias.h
+++ b/snapd-qt/Snapd/alias.h
@@ -14,7 +14,9 @@
 #include <Snapd/Enums>
 #include <Snapd/WrappedObject>
 
-class Q_DECL_EXPORT QSnapdAlias : public QSnapdWrappedObject {
+#include "snapdqt_global.h"
+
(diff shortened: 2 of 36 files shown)


aleasto:
```
#if defined(LIBSNAPDQT)
#define LIBSNAPDQT_EXPORT __attribute__((visibility("default")))
#else
#define LIBSNAPDQT_EXPORT Q_DECL_IMPORT
#endif
```

This boilerplate is usually specified in one dedicated header file which is then included by all the other headers that need it

sergio-costas:
@aleasto Very good point, indeed. I'll change it. But, of course, if you know a better way of doing this... 😇 (I still think of this as a very crude hack...)

snapd-qt/Snapd/qt-visibility.h
```diff
@@ -0,0 +1 @@
+#include "qt-visibility.h"
```
aleasto: Too much sed? 😅

aleasto (commented):
The canonical filename for such a header would be `snapdqt_global.h`.

snapd-qt/Snapd/qt-visibility.h
sergio-costas: yes 😅 Fixed

snapd-qt/Snapd/qt-visibility.h
```diff
@@ -0,0 +1,5 @@
+#if defined(LIBSNAPDQT)
```
aleasto, line 1: `#include <QtCore/QtGlobal>` for `Q_DECL_IMPORT`

sergio-costas: Done. Thanks!

sergio-costas:
@aleasto Thanks!

Issue #200, opened 2025-02-26
gcc compiler warning due to `const enum` return value


mcrha:
Compiling a user of the snapd-glib with gcc 14.2.1 I see this warning from the compiler:

```
In file included from /usr/include/snapd-glib-2/snapd-glib/snapd-client.h:28,
                 from /usr/include/snapd-glib-2/snapd-glib/snapd-glib.h:22,
/usr/include/snapd-glib-2/snapd-glib/snapd-notice.h:37:1: warning: type qualifiers ignored on function return type [-Wignored-qualifiers]
   37 | const SnapdNoticeType  snapd_notice_get_notice_type    (SnapdNotice *notice);
      | ^~~~~
```

meaning, as I understand it, adding `const` on enum/numeric return types is useless, thus ignored by the compiler.

A fix is to remove the `const` from the line.

sergio-costas:
Good point!

PR #201, opened 2025-02-26
Fix warning

sergio-costas:
The snapd_notice_get_notice_type() method returns a notice type, which is an `enum`. But the return type is defined as `const`, which causes a warning in the compiler when using the option `-Wignored-qualifiers`.

This patch removes the unneeded `const` qualifier to remove this warning.

Fix https://github.com/canonical/snapd-glib/issues/200

### DIFF
--- a/snapd-glib/snapd-notice.c
+++ b/snapd-glib/snapd-notice.c
@@ -103,7 +103,7 @@ const gchar *snapd_notice_get_user_id(SnapdNotice *self) {
  *
  * Since: 1.65
  */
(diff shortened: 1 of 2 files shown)


3v1n0:
Isn't this an API change though?

sergio-costas:
(SnapdNoticeType is an `enum`, not a pointer...)

robert-ancell (approved):
Looks good, I don't think this should cause any API issues, it's only reducing a requirement so existing code should still work. In general, there shouldn't be any use of const for the enum values.

PR #202, opened 2025-02-26
Remove support for Ubuntu 18.04

sergio-costas:
The checkout@v3 action relies on node20, which requires a version of GLibc newer than the one distributed in Ubuntu 18.04. Until now we used a workaround to be able to still run the action and, thus, execute the tests, but that workaround ceased to work and now neither the current code, nor any new PR do pass the continuous integration tests in github.

Since Ubuntu 18.04 reached end-of-life in May 2023, it makes sense to remove support.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -29,26 +29,12 @@ jobs:
             soup: libsoup2
             qt: qt5
             doc: false
-          - image: ubuntu:18.04
-            soup: libsoup2
-            qt: qt5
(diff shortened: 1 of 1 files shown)


sergio-costas:
@robert-ancell What do you think? Should we remove this?

robert-ancell (approved):
Agree it should be removed, this should make things easier to support. If any changes were required in 18.04, this would be done in the Ubuntu archives now.

PR #203, opened 2025-03-13
Use the new abstract socket inside a snap

sergio-costas:
The snap team has changed the socket path used to communicate with snapd from inside a snap to an abstract socket.

Although snapd-glib didn't manage the case of being used inside a snap, this change gives us the opportunity to both fix this, and ensure compatibility with the new socket and old systems.

This patch checks if the library is being run from inside a snap. If that's not the case, the old, classic socket /run/snapd.socket will be used. But if that's the case, the new abstract socket @/snapd/snapd-snap.socket will be tested first, and only if it's not possible to connect, will the old "inside snap" socket, /run/snapd-snap.socket, be used.


3v1n0 (approved):

3v1n0:
The only thing to note is that now technically this is blocking when used in async APIs, so ideally we should make the client being async initable instead

sergio-costas:
Not sure what do you mean... the open is just to test if the socket exists, it isn't being used for anything else... (in fact, it is closed as soon as `get_default_snap_socket()` returns...). If you mean that the `g_socket` is set to non-blocking, I don't think that affects this, because, again, it isn't used anywhere.

sergio-costas:
@3v1n0 Also, added a test to ensure that the socket test does work as expected.

meson_options.txt
```diff
@@ -25,5 +25,5 @@ option('examples',
        type: 'boolean', value: true,
        description: 'Whether to build the examples')
 option('tests',
-       type: 'boolean', value: true,
+       type: 'boolean', value: false,
```
3v1n0, line 28: Why false by default now?

snapd-glib/snapd-client.c
```diff
@@ -754,7 +756,13 @@ static GSocket *open_snapd_socket(const gchar *socket_path,
     return NULL;
   }
   g_socket_set_blocking(socket, FALSE);
-  g_autoptr(GSocketAddress) address = g_unix_socket_address_new(socket_path);
+  g_autoptr(GSocketAddress) address = NULL;
+  if (socket_path[0] == '@') {
```
3v1n0, line 760: maybe add a g_assert(socket_path != NULL) too?

snapd-glib/snapd-client.c
```diff
@@ -764,6 +772,17 @@ static GSocket *open_snapd_socket(const gchar *socket_path,
   return g_steal_pointer(&socket);
 }
 
+#ifdef BUILD_TESTS
```
3v1n0, line 775: What about exposing the `open_snapd_socket` function in a private header instead?

So that we don't have to do the actual testing inside the main file

meson_options.txt
sergio-costas: Because the tests require access to some static code, so when compiling with the tests enabled, a new function is exported that just calls that static function. But in normal circumstances we don't want to have that function exported, so I decided to disable the tests by default.

snapd-glib/snapd-client.c
sergio-costas: Mmm... good point...

meson_options.txt
sergio-costas: Now that I used a private header, this change isn't needed anymore. Reverted.

snapd-glib/snapd-client.c
sergio-costas: Added a `g_return_val_if_fail()`.

snapd-glib/snapd-client.c
sergio-costas: Done.

sergio-costas:
There's a problem with libsoup2... I'll check it with valgrind.

sergio-costas:
@3v1n0 Ok, the tests do pass. It should be ready for another review.

snapd-glib/snapd-client-private.h
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+#pragma once
+
+/*
+ * This function is declared in this private header because it must be private,
+ * but also is needed in the tests.
+ */
+
+static GSocket *open_snapd_socket(const gchar *socket_path,
```
robert-ancell, line 17: There's no need to put the whole function in the header, just remove the static from the .c file and put the function prototype here. The function won't be made public in the library as it doesn't start with `snapd_` (see `snapd-glib/snapd-glib.map`.

However, I would just copy this code into the tests. There's no real reason to make it shared.

snapd-glib/snapd-client.c
```diff
@@ -1007,10 +989,27 @@ gboolean snapd_client_connect_finish(SnapdClient *self, GAsyncResult *result,
   return g_task_propagate_boolean(G_TASK(result), error);
 }
 
+static const gchar *get_default_snapd_socket() {
```
robert-ancell, line 992: This makes an unnecessary connect the first time the client is created. I think it would be better to keep connecting on demand.

The current logic is:
```
client.path = default_path
func set_path (path)
    client.path = path
func connect ()
   connect_socket (client.path)
```

I think the new logic should be:
```
client.path = null
func set_path (path)
    client.path = path
func connect ()
    if client.path != null
        connect_socket (client.path)
    else
        if !connect_socket (abstract_path)
            connect_socket (default_path)
```

snapd-glib/snapd-client-private.h
sergio-costas: That's what I tried before, but didn't work because the tests seem to link with the shared library instead of using the object files (during compilation it didn't find the function). That's why first I added an extra function that was published only when building the tests and putting the declaration directly in the test code, and then followed the idea from Marco of just putting the code in a private header.

About copying this code into the tests, I disagree, because any change done in the library would require remembering to do the same change in the tests, due to the code duplication.

snapd-glib/snapd-client.c
sergio-costas: But currently, `snapd_client_get_socket_path()` returns the socket path that will be used even if it hasn't been set before. Doing that would change the behavior, returning NULL if the user didn't set the path. Of course, we can either change that behavior, or we can add a check to know if we are inside a SNAP (although in that case we won't be able to know if we will use the abstract socket or the old one).

robert-ancell: If the socket path is null, `snapd_client_get_socket_path` could return the first path that will be attempted.

snapd-glib/snapd-client-private.h
robert-ancell: Looking at the test in more detail shows why - it doesn't make sense to be testing an internal part of the SnapdClient logic. The tests should modify MockSnapd to support an abstract socket and then do something simple like `snapd_client_get_system_information_sync` and confirm it works.

snapd-glib/snapd-client.c
sergio-costas: ok.

sergio-costas: Done.

sergio-costas:
@robert-ancell Ok, this should be ready. BTW, when this is merged, I'll tag a new version, since this change is required for next versions of snapd.

robert-ancell (approved):
LGTM, thanks!

PR #204, opened 2025-03-26
Update version number

sergio-costas:
After setting the 1.67 tag, it's required to update the version value to 1.68 in meson.build.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,5 +1,5 @@
 project ('snapd-glib', [ 'c', 'cpp' ],
-         version: '1.67',
+         version: '1.68',
          meson_version: '>= 0.58.0',
          default_options : [ 'c_std=c11' ])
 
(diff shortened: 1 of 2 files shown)


sergio-costas:
@robert-ancell Just to be sure that I'm not forgetting anything.

robert-ancell (changes requested):
Needs other files to be updated, see previous commit 94364fc6a7adf298588264290a81215e2f281e36 for example.

sergio-costas:
@robert-ancell I checked it, but the `snapd-glib-sections.txt` file in the `doc/reference` folder doesn't exist anymore since the change to `gi-docgen`.

robert-ancell:
That's fine, but it should be updating `snapd-glib/snapd-version.h`

sergio-costas:
Ops... My fault, I did the change but forgot to do the push...

robert-ancell (approved):
LGTM

PR #205, opened 2025-04-07
Install the globals file, otherwise we can't build the rest of headers

aleixpol:
Might have missed something, not super familiar with meson.


sergio-costas (approved):
Absolutely right. Sorry.

PR #212, opened 2025-06-05
Fix build in ARCH

sergio-costas:
It's not possible to use Q_DECL_EXPORT due to Arch compiling with no_direct_extern_access (https://github.com/canonical/snapd-glib/issues/175 ) so our own macro LIBSNAPDQT_EXPORT must be used.

This patch fixes it.


aleasto (approved):
Yep

Issue #211, opened 2025-06-04
Build error for 1.69


username227:
Receiving the following build error on Arch Linux when building latest release 1.69:
```
[218/219] Linking target tests/test-qt6
FAILED: tests/test-qt6 
c++  -o tests/test-qt6 tests/test-qt6.p/meson-generated_moc_test-qt.cpp.o tests/test-qt6.p/test-qt.cpp.o -Wl,--as-needed -Wl,--no-undefined -Wl,-O1 -Wl,-O1 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,-z,pack-relative-relocs -flto=auto -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fexceptions -Wp,-D_FORTIFY_SOURCE=3 -Wformat -Werror=format-security -fstack-clash-protection -fcf-protection -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -Wp,-D_GLIBCXX_ASSERTIONS -g -ffile-prefix-map=/build/snapd-glib/src=/usr/src/debug/snapd-glib -flto=auto '-Wl,-rpath,$ORIGIN/../snapd-qt:$ORIGIN/../snapd-glib' -Wl,--start-group tests/libmock-snapd.a snapd-qt/libsnapd-qt-2.so.1.0.0 /usr/lib/libglib-2.0.so /usr/lib/libQt6Core.so /usr/lib/libQt6Network.so /usr/lib/libjson-glib-1.0.so /usr/lib/libgio-2.0.so /usr/lib/libgobject-2.0.so /usr/lib/libsoup-3.0.so /usr/lib/libgmodule-2.0.so -Wl,--end-group -pthread
/usr/bin/ld: /tmp/ccIMNbRu.ltrans6.ltrans.o: in function `test_find_query() [clone .lto_priv.0]':
/usr/src/debug/snapd-glib/snapd-glib-1.69/build/../tests/test-qt.cpp:3728:(.text+0x4f7d): undefined reference to `QSnapdLink::type() const'
/usr/bin/ld: /usr/src/debug/snapd-glib/snapd-glib-1.69/build/../tests/test-qt.cpp:3729:(.text+0x501f): undefined reference to `QSnapdLink::urls() const'
[...]

PR #213, opened 2025-06-05
Ensure that Q_DECL_EXPORT is not used

sergio-costas:
As seen in https://github.com/canonical/snapd-glib/issues/175 , using Q_DECL_EXPORT can result in not being able to build in Arch.

This test ensures that no Q_DECL_EXPORT is used, and explains why and how to replace it with the right macro.


sergio-costas:
@aleasto This is a test to ensure that https://github.com/canonical/snapd-glib/issues/211 doesn't happen again.

PR #215, opened 2025-07-16
ci: Allow to rename the default branch to `main`

sergio-costas:
Currently, the default branch is `master`, but we are renaming all of them to the current default name `main`. Unfortunately, the CI tests expects the default branch to be `master`.

This patch changes this to support both branches, thus allowing the renaming at any moment.

### DIFF
--- a/.github/workflows/check_q_decl_export.yaml
+++ b/.github/workflows/check_q_decl_export.yaml
@@ -4,6 +4,7 @@ on:
   pull_request:
   push:
     branches:
+      - main
       - master
 
 jobs:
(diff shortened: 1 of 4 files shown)


robert-ancell:
Historical note: master was used to match snapd, which is still using this.

sergio-costas:
> Historical note: master was used to match snapd, which is [still using this](https://bugs.launchpad.net/snapd/+bug/1995369).

Oh... Another example of the Chesterton's fence...? Can this change mean any problem for snapd-glib?

Issue #216, opened 2025-10-30
Qt5 QML module built against libsoup3 fails to load


fredldotme:
System: Ubuntu Touch 24.04-2.x

The QML module is, as per meson.build, appending a "2" to the resulting directory and file names, which breaks QMLs ability to look up the expected qmldir.

Renaming "Snapd2" to "Snapd" for both the directory and the library fixes the issue.

fredldotme:
We had a discussion internally and we might have figured out the rationale behind also renaming the QML path.

If you want I can propose a separate PR that makes libsoup3 builds require `import Snapd2 1.0` from QML. However you prefer.

sergio-costas:
I agree: soup2 and soup3 libraries can't be mixed in the same code, so it's a good idea to be able to choose to which Snapd version a project is compiled against, and be able to have both installed at the same time.

sergio-costas:
Also, sorry for the delay looking this. We are at the semestral sprint and have very little spare time.

PR #218, opened 2025-11-04
snapd-qt: Change soup3-based QML module name to Snapd2

fredldotme:
Force QML binding consumers to select either `Snapd` or `Snapd2` as the module import for libsoup2 and libsoup3 not to clash in the same process.

For libsoup2 builds require: `import Snapd 1.0`
For libsoup3 builds require: `import Snapd2 1.0`

Fixes: https://github.com/canonical/snapd-glib/issues/216

### DIFF
--- a/snapd-qt/meson.build
+++ b/snapd-qt/meson.build
@@ -185,9 +185,6 @@ pc.generate (libraries: snapd_qt_lib,
               version: meson.project_version (),
               requires: pc_required_qt,)
 
-install_data ('qmldir',
-              install_dir: qml_dir)
-
 cmake_conf = configuration_data ()
 cmake_conf.set ('libdir', libdir)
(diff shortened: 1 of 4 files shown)


sergio-costas:
This patch looks good. @robert-ancell Opinions?

fredldotme:
I am unsure what the CLA service is doing, I have signed the CLA multiple times now..

sergio-costas:
Anyway, there is the other problem with the signature...

fredldotme:
Just an FYI for when you package this: In Ubuntu Touch we renamed the module package to `qml-module-snapd2` as per https://alioth-lists.debian.net/pipermail/pkg-kde-talk/2014-March/001889.html. You might want to do the same.

sergio-costas:
@fredldotme Should I merge this too?

PR #219, opened 2025-11-08
qml-plugin: Register QSnapdGetSnapsRequest type in QML

fredldotme:
Otherwise SnapdClient.getSnaps() won't pass through.


fredldotme:
> Could you please run the `format-source.sh` script to format the changes?

Done.

PR #222, opened 2026-02-17
client: Use Cancellable to control the read source using independent requests

3v1n0:
We need to cancel the source when the request got cancelled


3v1n0:
> But what about API/ABI compatibility? Does it change anything? (it doesn't seem to do, but just in case...)

No, of course no, these are all private (static) functions.

PR #224, opened 2026-02-25
Add ask_interface_request client method for permission prompting

quine00:
Add a new POST /v2/interfaces/requests request type that allows callers to ask snapd whether a snap should be granted access to a given interface. The request carries the interface name, uid, pid and snap label; the response contains an "outcome" object with "allow"/"deny" booleans indicating the conclusion of the prompt.

This will be used initially by WirePlumber's
audio-permission-prompting module to query snapd before granting a snap access to audio record/playback.

This is in preparation for https://github.com/canonical/snapd/pull/16630


sergio-costas (commented):
Looks good, but there are some formatting problems...

snapd-glib/requests/snapd-post-interfaces-requests.c
```diff
@@ -0,0 +1,106 @@
...
+          "ready-callback", callback, "ready-callback-data", user_data, NULL));
+
+  self->interface = g_strdup(interface);
+  self->pid = pid;
+
+  return self;
+}
+
+static SoupMessage *
+generate_post_interfaces_requests_request(SnapdRequest *request,
```
olivercalder: Arguably it could be nice to pass in `"action"` as well, since theoretically we could in the future have other actions. But then `_snapd_post_interfaces_requests_new` would probably not want to make interface and pid required parameters.

This is fine either way, I don't know how other endpoints with different actions handle this. And I have a hard time following how all these special glib functions and macros interrelate 😅

olivercalder (commented):
This looks correct to me, though I have several questions since I'm really not familiar with glib/gobject code and it's difficult for me to follow how things relate. The primary things I want to make sure are handled correctly are errors from snapd (namely 500 and 503), and that the sync/async functions make sense, since it's all fully blocking on the snapd side.

snapd-glib/requests/snapd-post-interfaces-requests.c
quine00: Fair point, the public API is already action-specific, so it's a stable boundary if the internals need to evolve for new request types. I kept it like this for simplicity, if in the future new types of requests come in, it will be easier to generalise the internals once the shape is known imo

snapd-glib/snapd-client.c
```diff
@@ -4959,6 +4960,61 @@ gchar *snapd_client_get_serial_assertion_finish(SnapdClient *self,
...
+gboolean snapd_client_ask_interface_request_finish(SnapdClient *self,
+                                                   GAsyncResult *result,
+                                                   GError **error) {
+  g_return_val_if_fail(SNAPD_IS_CLIENT(self), FALSE);
+  g_return_val_if_fail(SNAPD_IS_POST_INTERFACES_REQUESTS(result), FALSE);
+
+  SnapdPostInterfacesRequests *request = SNAPD_POST_INTERFACES_REQUESTS(result);
+
+  if (!_snapd_request_propagate_error(SNAPD_REQUEST(request), error))
+    return FALSE;
```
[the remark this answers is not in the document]
quine00: If there's an error and the function returns false, the details are propagated in the GError, which the caller can use to decide on how to handle the error.  For a 503, Wireplumber for instance will enter a retry loop.

quine00:
> This looks correct to me, though I have several questions since I'm really not familiar with glib/gobject code and it's difficult for me to follow how things relate. The primary things I want to make sure are handled correctly are errors from snapd (namely 500 and 503), and that the sync/async functions make sense, since it's all fully blocking on the snapd side.

Thanks for the review Oliver, updated now with the test cases addressed.

snapd-glib/requests/snapd-post-interfaces-requests.c
olivercalder: Good reasoning, agreed.

olivercalder (approved):
Thanks for the explanations, looks good to me!

snapd-glib/requests/snapd-post-interfaces-requests.c
```diff
@@ -0,0 +1,106 @@
...
+}
+
+static gboolean parse_post_interfaces_requests_response(
+    SnapdRequest *request, guint status_code, const gchar *content_type,
+    GBytes *body, SnapdMaintenance **maintenance, GError **error) {
+  SnapdPostInterfacesRequests *self = SNAPD_POST_INTERFACES_REQUESTS(request);
+
+  g_autoptr(JsonObject) response =
+      _snapd_json_parse_response(content_type, body, maintenance, NULL, error);
+  if (response == NULL)
```
nteodosio: There is no logging facility in Snapd-Glib, is there? Just wondering if it could be useful to have something here to discern no response from invalid JSON.

nteodosio (approved):

3v1n0 (commented):
Looks good from upstream POV, a part some minor comments.

Now, regarding downstream: do we have an FFe for this (if this is 26.04 material)?

snapd-glib/snapd-client.c
```diff
@@ -4931,6 +4932,54 @@ gchar *snapd_client_get_serial_assertion_finish(SnapdClient *self,
...
+
+  g_autoptr(SnapdPostInterfacesRequests) request =
+      _snapd_post_interfaces_requests_new(interface, pid, cancellable, callback,
+                                          user_data);
+  send_request(self, SNAPD_REQUEST(request));
+}
+
+/**
+ * snapd_client_ask_interface_request_finish:
+ * @client: a #SnapdClient.
```
3v1n0, line 4964: This is wrong (`self` not client), but also make sure that he variable name is matching header and c file

snapd-glib/snapd-client.c
```diff
@@ -4931,6 +4932,54 @@ gchar *snapd_client_get_serial_assertion_finish(SnapdClient *self,
...
+ * Complete request started with snapd_client_ask_interface_request_async().
+ * See snapd_client_ask_interface_request_sync() for more information.
+ *
+ * Returns: %TRUE if the snap should be granted access, %FALSE otherwise.
+ *
+ * Since: 1.72
+ */
+gboolean snapd_client_ask_interface_request_finish(SnapdClient *self,
+                                                   GAsyncResult *result,
+                                                   GError **error) {
```
3v1n0, line 4977: These parameters must be mentioned in docs, what is not needed, is their details  (transfer or other metdata)

snapd-glib/snapd-client-sync.c
```diff
@@ -2119,3 +2119,32 @@ gchar *snapd_client_get_serial_assertion_sync(SnapdClient *self,
...
+ *
+ * Returns: %TRUE if the snap should be granted access, %FALSE otherwise.
+ *
+ * Since: 1.72
+ */
+gboolean snapd_client_ask_interface_request_sync(SnapdClient *self,
+                                                 const gchar *interface,
+                                                 GPid pid,
+                                                 GCancellable *cancellable,
+                                                 GError **error) {
```
3v1n0, line 2142: cancellable and error are missing in the docstring

snapd-glib/snapd-client-sync.c
```diff
@@ -2119,3 +2119,32 @@ gchar *snapd_client_get_serial_assertion_sync(SnapdClient *self,
   end_sync(&data);
   return snapd_client_get_serial_assertion_finish(self, data.result, error);
 }
+
+/**
+ * snapd_client_ask_interface_request_sync:
+ * @client: a #SnapdClient.
```
3v1n0, line 2125: Should be `self`. And make sure header and c file variable name match (or just use `client` everywhere)

